PR #17, opened 2023-11-21
crypto crate implementation

amika-sq:
This PR is a first pass at an implementation for the `crypto` crate. 

This implementation utilized the `ssi_jwk` and `ssi_jws` crates from Spruce's excellent SSI library: https://github.com/spruceid/ssi/

The structure & operations here follows that in [web5-kt/crypto](https://github.com/TBD54566975/web5-kt/tree/main/crypto) with a few notable exceptions, which I have outlined below. I am open to changing/renaming any of these!

## `PrivateKey` and `PublicKey`
There is an additional abstraction over JWKs: `PrivateKey` and `PublicKey`. This allows us to move signing / verifying closer to the actual objects doing the work. 

Spruce's ssi library utilizes JWKs as they key type for most of it's operations, so these abstractions directly wrap a JWK. In the future, if we want to introduce another key type, we can convert these into generic traits.

This abstraction also helps with the `uniffi-rs` binding generation process which we will be using for Swift, which will be coming in a later PR. Because we don't own the `JWK` type (it's defined in `ssi_jwk`), we can't directly expose it to the binding generator without duplicating it. Using Rust newtypes that wrap the JWK allows us to get around this.

## `LocalKeyManager` and `KeyStore`

Our ultimate goal here is to export this Rust code into Swift code. On Swift, we want a way to generate cryptographic keys via Rust, but then store them using secure storage mechanisms such as the Keychain and the Secure Enclave. In order to accomodate this, I needed to split apart the "key generation" and "key storage" tasks into separate traits.

`LocalKeyManager` is how I did this. It's responsibilities are to generate the cryptographic keys locally (locally meaning in Rust) and then delegate the actual storage to a `KeyStore` trait. 

I have provided a `InMemoryKeyStore` implementation that stores the cryptographic keys in memory. But anyone can provide their own implementation, including foreign languages.

### DIFF
--- a/crates/crypto/Cargo.toml
+++ b/crates/crypto/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "crypto"
+version = "0.1.0"
+edition = "2021"
+homepage.workspace = true
+repository.workspace = true
+license-file.workspace = true
+
+[dependencies]
+ssi-jwk = "0.1.2"
+ssi-jws = "0.1.1"
+thiserror = { workspace = true }
--- a/crates/crypto/src/key/key.rs
+++ b/crates/crypto/src/key/key.rs
@@ -1,0 +1,22 @@
+use ssi_jwk::JWK;
+use ssi_jws::Error as JWSError;
+
+/// Enum defining all supported cryptographic algorithms.
+pub enum KeyAlgorithm {
+    Secp256k1,
+    Secp256r1,
+    Ed25519,
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum KeyError {
+    #[error(transparent)]
+    JWSError(#[from] JWSError),
+    #[error("Algorithm not found on JWK")]
+    AlgorithmNotFound,
+}
+
+/// Trait defining all common behavior for cryptographic keys.
+pub trait Key {
+    fn jwk(&self) -> &JWK;
+}
--- a/crates/crypto/src/key/private_key.rs
+++ b/crates/crypto/src/key/private_key.rs
@@ -1,0 +1,59 @@
+use crate::key::{Key, KeyError, PublicKey};
+use ssi_jwk::JWK;
+use ssi_jws::sign_bytes;
+
+#[derive(Clone, PartialEq, Debug)]
+pub struct PrivateKey(pub(crate) JWK);
+
+impl PrivateKey {
+    /// Derive a [`PublicKey`] from the target [`PrivateKey`].
+    pub fn to_public(&self) -> PublicKey {
+        PublicKey(self.0.to_public())
+    }
+
+    /// Sign a payload using the target [`PrivateKey`].
+    pub fn sign(&self, payload: &[u8]) -> Result<Vec<u8>, KeyError> {
+        let algorithm = self.0.get_algorithm().ok_or(KeyError::AlgorithmNotFound)?;
+        let signed_bytes = sign_bytes(algorithm, &payload, &self.0)?;
+
+        Ok(signed_bytes)
+    }
+}
+
+impl Key for PrivateKey {
+    fn jwk(&self) -> &JWK {
+        &self.0
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use ssi_jwk::JWK;
+
+    fn new_private_key() -> PrivateKey {
+        PrivateKey(JWK::generate_secp256k1().unwrap())
+    }
+
+    #[test]
+    fn test_to_public() {
+        let private_key = new_private_key();
+        let public_key = private_key.to_public();
+
+        assert_eq!(
+            private_key.jwk().thumbprint().unwrap(),
+            public_key.jwk().thumbprint().unwrap()
+        );
+
+        assert_ne!(private_key.jwk(), public_key.jwk())
+    }
+
+    #[test]
+    fn test_sign() {
+        let private_key = new_private_key();
+        let payload: &[u8] = b"hello world";
+        let signature = private_key.sign(payload).unwrap();
+
+        assert_ne!(payload, &signature)
+    }
+}
--- a/crates/crypto/src/key_manager/key_store/in_memory_key_store.rs
+++ b/crates/crypto/src/key_manager/key_store/in_memory_key_store.rs
@@ -1,0 +1,71 @@
+use crate::key::PrivateKey;
+use crate::key_manager::key_store::{KeyStore, KeyStoreError};
+use std::collections::HashMap;
+use std::sync::Mutex;
+
+/// An in-memory implementation of the [`KeyStore`] trait.
+pub struct InMemoryKeyStore {
+    map: Mutex<HashMap<String, PrivateKey>>,
+}
+
+impl InMemoryKeyStore {
+    pub fn new() -> Self {
+        let map = Mutex::new(HashMap::new());
+        Self { map }
+    }
+}
+
+impl KeyStore for InMemoryKeyStore {
+    fn get(&self, key_alias: &str) -> Result<Option<PrivateKey>, KeyStoreError> {
+        let map_lock = self.map.lock().map_err(|e| {
+            KeyStoreError::InternalKeyStoreError(format!("Unable to acquire Mutex lock: {}", e))
+        })?;
+
+        if let Some(private_key) = map_lock.get(key_alias) {
+            Ok(Some(private_key.clone()))
+        } else {
+            Ok(None)
+        }
+    }
+
+    fn insert(&self, key_alias: &str, private_key: PrivateKey) -> Result<(), KeyStoreError> {
+        let mut map_lock = self.map.lock().map_err(|e| {
+            KeyStoreError::InternalKeyStoreError(format!("Unable to acquire Mutex lock: {}", e))
+        })?;
+
+        map_lock.insert(key_alias.to_string(), private_key);
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::key::PrivateKey;
+    use ssi_jwk::JWK;
+
+    fn new_private_key() -> PrivateKey {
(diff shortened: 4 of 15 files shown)


crates/crypto/src/key/key.rs
```diff
@@ -0,0 +1,22 @@
+use ssi_jwk::JWK;
+use ssi_jws::Error as JWSError;
+
+/// Enum defining all supported cryptographic algorithms.
+pub enum KeyAlgorithm {
```
decentralgabe, line 5: nit: `KeyType` instead of algorithm

crates/crypto/src/key_manager/local_key_manager.rs
```diff
@@ -0,0 +1,133 @@
...
+        }
+    }
+
+    fn sign(&self, key_alias: &str, payload: &[u8]) -> Result<Vec<u8>, KeyManagerError> {
+        let private_key = self
+            .key_store
+            .get(key_alias)?
+            .ok_or(KeyManagerError::SigningKeyNotFound)?;
+
+        let signed_payload = private_key.sign(&payload.to_vec())?;
```
decentralgabe, line 59: is it assumed that all key types expose this sign method?

amika-sq: That's correct. The `PrivateKey` struct is currently responsible for all key types, which itself is a thin wrapper around a JWK. 

Definitely flexible here to alternative approaches!

decentralgabe: I think it may be worth including a signing algorithm parameter, as sometimes the same key can be used with different algorithms (e.g. secp256k1 can be used with ECDH and ECDSA)

crates/crypto/src/key_manager/key_store/in_memory_key_store.rs
```diff
@@ -0,0 +1,71 @@
+use crate::key::PrivateKey;
+use crate::key_manager::key_store::{KeyStore, KeyStoreError};
+use std::collections::HashMap;
+use std::sync::Mutex;
+
+/// An in-memory implementation of the [`KeyStore`] trait.
+pub struct InMemoryKeyStore {
+    map: Mutex<HashMap<String, PrivateKey>>,
+}
```
ethan-tbd, line 9: not very familiar w/ rust.. why is `Mutex` used here over something like `RwLock`?

amika-sq: This is a really good question.

* `Mutex` allows mutually exclusive access to the shared data, no matter if the caller is accessing the data for a read or write. Both read and writes will block the current thread until it acquires the lock. This is why there's only a singular `lock()` function to acquire a lock into the underlying data.
* `RwLock` allows for any number of readers, but requires exclusive access for writes. This means that a read lock will only block the current thread if there's an active write lock. This is why there's distinctive `read()` and `write()` functions to acquire a lock to the underlying data.

This makes `RwLock` preferable in read-heavy situations, where as `Mutex` is more appropriate for write-heavy situations.

In our case, writing keys to the store is going to be FAR less frequent than reading them. So we actually _should_ be using a `RwLock` here for better performance. I'll go ahead and update!

Thank you for asking this! Learning together!

Docs:
* [Mutex](https://doc.rust-lang.org/std/sync/struct.Mutex.html)
* [RwLock](https://doc.rust-lang.org/std/sync/struct.RwLock.html)

andresuribe87 (commented):
Forgot to hit send a bit back...

mistermoe (approved):
🏅

crates/crypto/src/key_manager/local_key_manager.rs
```diff
@@ -0,0 +1,127 @@
...
+use crate::key_manager::key_store::{InMemoryKeyStore, KeyStore};
+use crate::key_manager::{KeyManager, KeyManagerError};
+use ssi_jwk::JWK;
+use std::sync::Arc;
+
+/// Implementation of the [`KeyManager`] trait with key generation local to the device/platform it
+/// is being run. Key storage is provided by a [`KeyStore`] trait implementation, allowing the keys
+/// to be stored wherever is most appropriate for the application.
+pub struct LocalKeyManager {
+    key_store: Arc<dyn KeyStore>,
```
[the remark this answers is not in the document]
amika-sq: Arc is a shared data reference that uses the concept of reference counting. When you create an Arc via `Arc::new()` around some piece of data, that data is placed on the heap along with two counters for "strong" and "weak" references to it. After `Arc::new()`, it looks something like this:

Untitled scene

The `Arc` you get back is actually a pointer to some heap allocated memory (called `ArcInner`), which contains the counts, as well as the actual shared data itself. If the `Arc` pointer you get back goes out of scope and gets deallocated, the strong reference counter that the `ArcInner` was pointing to is atomically decremented. When the strong count falls to zero, the heap allocated memory that contains the actual data is then cleaned up.

Now looking at these particular lines:

```
pub struct LocalKeyManager {
    key_store: Arc<dyn KeyStore>,
}
```

What this is saying is that `LocalKeyManager` owns an `Arc`, which points to a KeyStore. The lifetime of the Arc will only end when the `LocalKeyManager`'s lifetime ends. This means that the strong reference count will remain non-zero for the duration of `LocalKeyManager`'s lifetime, guaranteeing that it's always available to use.

In order to construct a `LocalKeyManager`, the caller must to provide an `Arc<dyn Keystore>` value in the constructor. This can be done either by creating a new Arc like I mentioned earlier with `Arc::new()`, or the caller can clone an existing Arc.

Cloning an Arc isn't actually cloning the heap allocated data. Instead, it's cloning the pointer to that heap allocated data. As a part of the clone operation, the strong reference count is atomically incremented. So, even if the originally cloned Arc goes out of scope and gets deallocated, decrementing the reference count, the reference count is still non-zero.

---

I hope that made some sort of sense 😅 . Happy to chat through this more if you'd like more clarity, or had additional questions! Arc is definitely one of the more interesting data types in Rust, and it's heavily used throughout the bindings generation process which will come in future PRs.

crates/crypto/src/key_manager/local_key_manager.rs
amika-sq: This should still be possible, by the caller setting the algorithm field on the JWK. 

Currently, the algorithm is computed by this function in Spruce's `ssi-jwk` crate:
https://github.com/spruceid/ssi/blob/main/ssi-jwk/src/lib.rs#L341

This logic matches our signer lookup that's currently being done in Kotlin here:
https://github.com/TBD54566975/web5-kt/blob/main/crypto/src/main/kotlin/web5/sdk/crypto/Crypto.kt#L111C22-L111C22

crates/crypto/src/key_manager/local_key_manager.rs
```diff
@@ -0,0 +1,127 @@
...
+}
+
+impl LocalKeyManager {
+    /// Constructs a new `LocalKeyManager` that stores keys in the provided `KeyStore`.
+    pub fn new(key_store: Arc<dyn KeyStore>) -> Self {
+        Self { key_store }
```
andresuribe87, line 17: Is this a place where you should be using `::clone` on the passed in `key_store`? If not, mind elaborating why?

crates/crypto/src/key_manager/mod.rs
```diff
@@ -0,0 +1,7 @@
+mod key_manager;
+pub use key_manager::*;
```
andresuribe87, line 2: Feels like wildcard is difficult to maintain. I typically recommend being explicit on what this is exposing. 

Tangentially, would it make sense to setup a linter? I briefly looked online and found https://github.com/rust-lang/rust-clippy

crates/crypto/src/key_manager/local_key_manager.rs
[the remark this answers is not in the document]
andresuribe87: Thanks! Asked more in the line below.

crates/crypto/src/key_manager/local_key_manager.rs
amika-sq: No, not exactly. The contract here says "I need an `Arc<dyn KeyStore>`." It's the caller's responsibility to provide that. Whether that's done by a `::clone()` or `::new()` is an implementation detail that the caller must handle. If they do not, it's a compilation error.

crates/crypto/src/key_manager/mod.rs
amika-sq: This is pretty common in Rust. It's just saying that I can use the `key_manager` module without specifying `key_manager` again.

Without this line, caller's use statements would look like this:
`use crypto::key_manager::key_manager::KeyManager`

But with this line, callers can reference it like so:
`use crypto::key_manager::KeyManager`

crates/crypto/src/key/private_key.rs
```diff
@@ -0,0 +1,59 @@
...
+use ssi_jwk::JWK;
+use ssi_jws::sign_bytes;
+
+#[derive(Clone, PartialEq, Debug)]
+pub struct PrivateKey(pub(crate) JWK);
+
+impl PrivateKey {
+    /// Derive a [`PublicKey`] from the target [`PrivateKey`].
+    pub fn to_public(&self) -> PublicKey {
+        PublicKey(self.0.to_public())
```
jiyoontbd, line 11: can you explain the `.0.` part? after installing some vscode rust extensions, it tells me this 0 is equivalent to JWK (which i guess is the 0th param passed into the struct on line 6 above), is that correct? and if it is, is calling `self.0.foo()` different than  calling `self.jwk().foo()` ?

crates/crypto/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "crypto"
+version = "0.1.0"
+edition = "2021"
```
andresuribe87, line 4: 2023?

andresuribe87 (approved):

PR #18, opened 2023-11-28
Credentials crate & PresentationDefinitionV2 struct definition

amika-sq:
`PresentationDefinitionV2` is needed for a [tbDEX Offering](https://github.com/TBD54566975/tbdex/tree/main/specs/protocol#offering). 

This PR adds a `PresentationDefinitionV2` definition to a new `credentials` crate in this repository, which will be consumed in the `tbdex-rs` repository to define an Offering.

I primarily followed what is currently in the [Kotlin repository here](https://github.com/TBD54566975/web5-kt/blob/main/credentials/src/main/kotlin/web5/sdk/credentials/PresentationDefinition.kt) as an example, and re-used the comments / tests from there.

### DIFF
--- a/crates/credentials/Cargo.toml
+++ b/crates/credentials/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "credentials"
+version = "0.1.0"
+edition = "2021"
+homepage.workspace = true
+repository.workspace = true
+license-file.workspace = true
+
+[dependencies]
+jsonschema = "0.17.1"
+serde = { version = "1.0.193", features = ["derive"] }
+serde_json = "1.0.108"
+serde_with = "3.4.0"
--- a/crates/credentials/src/pex/v2/presentation_definition.rs
+++ b/crates/credentials/src/pex/v2/presentation_definition.rs
@@ -14,13 +14,13 @@ use std::collections::HashMap;
 /// for more information.
 #[skip_serializing_none]
 #[derive(Debug, Default, Deserialize, PartialEq, Serialize)]
-pub struct PresentationDefinitionV2 {
+pub struct PresentationDefinition {
     pub id: String,
     pub name: Option<String>,
     pub purpose: Option<String>,
     pub format: Option<Format>,
     pub submission_requirements: Option<Vec<SubmissionRequirement>>,
-    pub input_descriptors: Vec<InputDescriptorV2>,
+    pub input_descriptors: Vec<InputDescriptor>,
     pub frame: Option<HashMap<String, JsonValue>>,
 }
 
@@ -30,12 +30,12 @@ pub struct PresentationDefinitionV2 {
 /// for more information.
 #[skip_serializing_none]
 #[derive(Debug, Default, Deserialize, PartialEq, Serialize)]
-pub struct InputDescriptorV2 {
+pub struct InputDescriptor {
     pub id: String,
     pub name: Option<String>,
     pub purpose: Option<String>,
     pub format: Option<Format>,
-    pub constraints: ConstraintsV2,
+    pub constraints: Constraints,
 }
 
 /// Represents constraints for an input descriptor.
@@ -45,8 +45,8 @@ pub struct InputDescriptorV2 {
 /// for more information.
 #[skip_serializing_none]
 #[derive(Debug, Default, Deserialize, PartialEq, Serialize)]
-pub struct ConstraintsV2 {
-    pub fields: Option<Vec<FieldV2>>,
+pub struct Constraints {
+    pub fields: Option<Vec<Field>>,
     pub limit_disclosure: Option<ConformantConsumerDisclosure>,
 }
 
@@ -57,7 +57,7 @@ pub struct ConstraintsV2 {
 /// for more information.
 #[skip_serializing_none]
 #[derive(Debug, Default, Deserialize, PartialEq, Serialize)]
-pub struct FieldV2 {
+pub struct Field {
     pub id: Option<String>,
     pub path: Vec<String>,
     pub purpose: Option<String>,
@@ -67,7 +67,7 @@ pub struct FieldV2 {
     pub optional: Option<bool>,
 }
 
-impl FieldV2 {
+impl Field {
     pub fn filter_schema(&self) -> Option<JSONSchema> {
         self.filter
             .as_ref()
@@ -151,15 +151,15 @@ mod tests {
 
     #[test]
     fn can_serialize() {
-        let pd = PresentationDefinitionV2 {
+        let pd = PresentationDefinition {
             id: "tests-pd-id".to_string(),
             name: "simple PD".to_string().into(),
             purpose: "pd for testing".to_string().into(),
-            input_descriptors: vec![InputDescriptorV2 {
+            input_descriptors: vec![InputDescriptor {
                 id: "whatever".to_string(),
                 purpose: "purpose".to_string().into(),
-                constraints: ConstraintsV2 {
-                    fields: vec![FieldV2 {
+                constraints: Constraints {
+                    fields: vec![Field {
                         id: "field-id".to_string().into(),
                         path: vec!["$.issuer".to_string()],
                         purpose: "purpose".to_string().into(),
@@ -191,16 +191,16 @@ mod tests {
             ..Default::default()
         };
 
-        let expected_input_descriptors = vec![InputDescriptorV2 {
+        let expected_input_descriptors = vec![InputDescriptor {
             id: "7b928839-f0b1-4237-893d-b27124b57952".to_string(),
-            constraints: ConstraintsV2 {
+            constraints: Constraints {
                 fields: Some(vec![
-                    FieldV2 {
+                    Field {
                         path: vec!["$.iss".to_string(), "$.vc.issuer".to_string()],
                         filter: Some(json!({"type": "string", "pattern": "^did:[^:]+:.+"})),
                         ..Default::default()
                     },
-                    FieldV2 {
+                    Field {
                         path: vec!["$.vc.type[*]".to_string(), "$.type[*]".to_string()],
                         filter: Some(json!({"type": "string", "const": "SanctionsCredential"})),
                         ..Default::default()
@@ -212,7 +212,7 @@ mod tests {
         }];
 
         let pd_string = load_json("tests/resources/pd_sanctions.json");
-        let deserialized_pd: PresentationDefinitionV2 = serde_json::from_str(&pd_string).unwrap();
+        let deserialized_pd: PresentationDefinition = serde_json::from_str(&pd_string).unwrap();
 
         assert_eq!(deserialized_pd.id, expected_id);
         assert_eq!(deserialized_pd.format, Some(expected_format));
--- a/crates/credentials/src/pex/v2/presentation_definition.rs
+++ b/crates/credentials/src/pex/v2/presentation_definition.rs
@@ -10,7 +10,7 @@ use std::collections::HashMap;
 /// can use to articulate proof requirements, and a Presentation Submission data format Holders can
 /// use to describe proofs submitted in accordance with them.
 ///
-/// See [Presentation Definition](https://identity.foundation/presentation-exchange/#presentation-definition)
+/// See [Presentation Definition](https://identity.foundation/presentation-exchange/spec/v2.0.0/#presentation-definition)
 /// for more information.
 #[skip_serializing_none]
 #[derive(Debug, Default, Deserialize, PartialEq, Serialize)]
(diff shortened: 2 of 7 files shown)


crates/credentials/src/presentation_definition_v2.rs
```diff
@@ -0,0 +1,230 @@
...
+                },
+                ..Default::default()
+            }],
+            ..Default::default()
+        };
+
+        let serialized_pd = serde_json::to_string(&pd).unwrap();
+
+        assert!(serialized_pd.contains("input_descriptors"));
+        assert!(serialized_pd.contains("123"));
```
amika-sq, line 180: I'm not really convinced that this test fully tests serialization, but it's currently what we're doing in the [web5-kt](https://github.com/TBD54566975/web5-kt/blob/main/credentials/src/test/kotlin/web5/sdk/credentials/PresentationDefinitionTest.kt#L51-L52) test suite.

We could really use some test vectors here to have a more robust test of this functionality.

crates/credentials/src/presentation_definition_v2.rs
```diff
@@ -0,0 +1,230 @@
...
+}
+
+/// Represents constraints for an input descriptor.
+///
+/// See 'constraints object' defined in
+/// [Input Descriptor](https://identity.foundation/presentation-exchange/#input-descriptor-object)
+/// for more information.
+#[skip_serializing_none]
+#[derive(Debug, Default, Deserialize, PartialEq, Serialize)]
+pub struct ConstraintsV2 {
```
decentralgabe, line 48: nit: do we have to suffix with V2 here and elsewhere? I don't believe we have any need/desire to support v1

decentralgabe (approved):
lgtm

crates/credentials/src/presentation_definition_v2.rs
amika-sq: I'm not opposed, but I do think it helps with clarity. There are two distinct versions of PresentationDefinition, and if v2 was omitted, I'd think this was v1 at first glance.

We currently have the "V2" suffix in the spec for offerings here: https://github.com/TBD54566975/tbdex/tree/main/specs/protocol#offering

The  Kotlin & JS codebases also have the "V2" suffix as well.

If we do decide to remove, we should probably align on that across our codebases & documentation.

andresuribe87: What about namespacing (directory or module)? That would be my vote across all codebases.

crates/credentials/src/presentation_definition_v2.rs
andresuribe87: I agree. I think most is already tested in the idempotency test.

crates/credentials/src/presentation_definition_v2.rs
```diff
@@ -0,0 +1,230 @@
...
+
+        let pd_string = load_json("tests/resources/pd_sanctions.json");
+        let deserialized_pd: PresentationDefinitionV2 = serde_json::from_str(&pd_string).unwrap();
+
+        assert_eq!(deserialized_pd.id, expected_id);
+        assert_eq!(deserialized_pd.format, Some(expected_format));
+        assert_eq!(
+            deserialized_pd.input_descriptors,
+            expected_input_descriptors
+        );
```
andresuribe87, line 222: I think this is the most powerful test. In kt, it checks that JSON is equal (see [serialize / deserialize idempotency](https://github.com/TBD54566975/web5-kt/blob/main/credentials/src/test/kotlin/web5/sdk/credentials/PresentationDefinitionTest.kt#L56)). Would it make sense to compare the the actual JSON outputs from the pre-serialization JSON vs the serialized and then deserialized JSON?

crates/credentials/src/presentation_definition_v2.rs
```diff
@@ -0,0 +1,230 @@
...
+/// Presentation Exchange
+///
+/// Presentation Exchange specification codifies a Presentation Definition data format Verifiers
+/// can use to articulate proof requirements, and a Presentation Submission data format Holders can
+/// use to describe proofs submitted in accordance with them.
+///
+/// See [Presentation Definition](https://identity.foundation/presentation-exchange/#presentation-definition)
+/// for more information.
+#[skip_serializing_none]
+#[derive(Debug, Default, Deserialize, PartialEq, Serialize)]
```
andresuribe87, line 16: Curious why the `Debug, Default, PartialEq` are needed?

crates/credentials/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "credentials"
+version = "0.1.0"
+edition = "2021"
```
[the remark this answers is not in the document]
amika-sq: It's editions. More info here:
https://github.com/TBD54566975/web5-rs/pull/17#discussion_r1406655556

crates/credentials/src/presentation_definition_v2.rs
amika-sq: I couldn't find a canonicalizer written in Rust that behaves the same way as the Kotlin test. Whitespace differences caused issues when comparing the json strings.

That said, I did vastly improve the deserialize test to ensure that the JSON is parsed into the expected format here, whereas Kotlin is just testing that it doesn't throw when deserializing. It _should_ be testing the same thing.

crates/credentials/src/presentation_definition_v2.rs
amika-sq: `derive` is generating additional code at compile time for this struct:
* `Debug` so I can print out the values and get a debug description of the struct & all of its fields
* `Default` so I can create this struct with pre-populated default values (used in tests)
* `PartialEq` so I can use the `==` operator in tests

crates/credentials/src/presentation_definition_v2.rs
amika-sq: That's a good idea. I'll update this to namespace these into a `v2` module. Then we can drop the suffix!

amika-sq: Updated to remove the `V2` suffixes, and placed it within a `v2` module: 
https://github.com/TBD54566975/web5-rs/pull/18/commits/1ddeb54df94706146055f2a883b1b7cb39a2853a

Now consumers can use this in their code with the following use statement:
`use credentials::pex::v2::PresentationDefinition;`

crates/credentials/src/pex/v2/presentation_definition.rs
```diff
@@ -0,0 +1,230 @@
...
+use serde_with::skip_serializing_none;
+use std::collections::HashMap;
+
+/// Presentation Exchange
+///
+/// Presentation Exchange specification codifies a Presentation Definition data format Verifiers
+/// can use to articulate proof requirements, and a Presentation Submission data format Holders can
+/// use to describe proofs submitted in accordance with them.
+///
+/// See [Presentation Definition](https://identity.foundation/presentation-exchange/#presentation-definition)
```
KendallWeihe, line 13: I appreciate the links to spec docs 👍

crates/credentials/src/pex/v2/presentation_definition.rs
```diff
@@ -0,0 +1,230 @@
...
+/// for more information.
+#[skip_serializing_none]
```
KendallWeihe, line 15: So what's the intent of [`skip_serialization_none`](https://docs.rs/serde_with/latest/serde_with/attr.skip_serializing_none.html)? 

Is it such that serialization/deserialization doesn't throw exceptions in the event of optional fields missing?

crates/credentials/src/pex/v2/presentation_definition.rs
```diff
@@ -0,0 +1,230 @@
...
+    pub optional: Option<bool>,
+}
+
+impl Field {
+    pub fn filter_schema(&self) -> Option<JSONSchema> {
+        self.filter
+            .as_ref()
+            .map(|json| {
+                JSONSchema::options()
+                    .with_draft(Draft::Draft7)
```
KendallWeihe, line 76: What is the concept of a "draft" here?

crates/credentials/src/pex/v2/presentation_definition.rs
decentralgabe:
```suggestion
/// See [Presentation Definition](https://identity.foundation/presentation-exchange/spec/v2.0.0/#presentation-definition)
```

crates/credentials/src/pex/v2/presentation_definition.rs
```diff
@@ -0,0 +1,230 @@
...
+    pub name: Option<String>,
+    pub purpose: Option<String>,
+    pub format: Option<Format>,
+    pub constraints: Constraints,
+}
+
+/// Represents constraints for an input descriptor.
+///
+/// See 'constraints object' defined in
+/// [Input Descriptor](https://identity.foundation/presentation-exchange/#input-descriptor-object)
```
decentralgabe, line 44:
```suggestion
/// [Input Descriptor](https://identity.foundation/presentation-exchange/spec/v2.0.0/#input-descriptor-object)
```

crates/credentials/src/pex/v2/presentation_definition.rs
```diff
@@ -0,0 +1,230 @@
...
+#[derive(Debug, Default, Deserialize, PartialEq, Serialize)]
+pub struct Constraints {
+    pub fields: Option<Vec<Field>>,
+    pub limit_disclosure: Option<ConformantConsumerDisclosure>,
+}
+
+/// Represents a field in a presentation input descriptor.
+///
+/// See 'fields object' as defined in
+/// [Input Descriptor](https://identity.foundation/presentation-exchange/#input-descriptor-object)
```
decentralgabe, line 56:
```suggestion
/// [Input Descriptor](https://identity.foundation/presentation-exchange/spec/v2.0.0/#input-descriptor-object)
```

crates/credentials/src/pex/v2/presentation_definition.rs
decentralgabe: it's a revision for json schema itself...unfortunately they're still using quite an old draft

crates/credentials/src/pex/v2/presentation_definition.rs
```diff
@@ -0,0 +1,230 @@
...
+                    .ok()
+            })
+            .flatten()
+    }
+}
+
+/// Enumeration representing consumer disclosure options.
+///
+/// Represents the possible values of `limit_disclosure' property as defined in
+// [Input Descriptor](https://identity.foundation/presentation-exchange/#input-descriptor-object)
```
decentralgabe, line 87:
```suggestion
// [Input Descriptor](https://identity.foundation/presentation-exchange/spec/v2.0.0/#input-descriptor-object)
```

crates/credentials/src/pex/v2/presentation_definition.rs
```diff
@@ -0,0 +1,230 @@
...
+#[serde(rename_all = "snake_case")]
+pub enum ConformantConsumerDisclosure {
+    Required,
+    Preferred,
+}
+
+/// Represents the format of a presentation definition
+///
+/// See `format` as defined in
+/// [Input Descriptor](https://identity.foundation/presentation-exchange/#input-descriptor-object)
```
decentralgabe, line 98:
```suggestion
/// [Input Descriptor](https://identity.foundation/presentation-exchange/spec/v2.0.0/#input-descriptor-object)
```

KendallWeihe (approved):

crates/credentials/src/pex/v2/presentation_definition.rs
andresuribe87: I believe it makes it so that when serializing a struct to JSON, fields who's value is set to `Option::None` will be skipped. If this isn't set, then the fields woulb be serialized as `null`.

A more explicit example below. 

Without the `skip_serializing_none` macro
```json
{
  "foo": 1
  "bar": null
}
```


With the macro

```json
{
  "foo": 1
}
```

crates/credentials/src/presentation_definition_v2.rs
andresuribe87: Curious if you took a look into https://docs.rs/serde_jcs/latest/serde_jcs/ ?

adammika: Yes. Multiple times. Works well, but I couldn't get it to match white space between json of just the string vs deserialized jcs. I don't know why yet. In all honesty, I don't think it's worth fretting over yet. I'll make a task in the backlog to figure it out.

andresuribe87: Sounds good! I'd love to take a look at that as a first stab at rust!

crates/credentials/src/pex/v2/presentation_definition.rs
KendallWeihe: Yup that makes sense, thanks!

PR #19, opened 2023-11-29
Idempotency test for serialization of presentation definitions

andresuribe87:
Modified a test so that it mimics what was done in `web5-kt` (see https://github.com/TBD54566975/web5-kt/blob/a644e6f129cabc85906099d8e79497a1d18fff71/credentials/src/test/kotlin/web5/sdk/credentials/PresentationDefinitionTest.kt#L56)

The idea behind this test is that it ensures that fields weren't misspelled, or missed, from the `PresentationDefinition` struct. It also showcases:
* how to use a dependency that's only for testing purposes. 
* How to do canonicalization of JSON.

### DIFF
--- a/crates/credentials/Cargo.toml
+++ b/crates/credentials/Cargo.toml
@@ -11,3 +11,6 @@ jsonschema = "0.17.1"
 serde = { version = "1.0.193", features = ["derive"] }
 serde_json = "1.0.108"
 serde_with = "3.4.0"
+
+[dev-dependencies]
+serde_canonical_json = "1.0.0"
--- a/crates/credentials/src/pex/v2/presentation_definition.rs
+++ b/crates/credentials/src/pex/v2/presentation_definition.rs
@@ -145,9 +145,11 @@ pub enum Optionality {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use serde_json::json;
+    use serde_json::{json, Value};
     use std::fs;
     use std::path::Path;
+    use serde_json::Serializer;
+    use serde_canonical_json::CanonicalFormatter;
 
     #[test]
     fn can_serialize() {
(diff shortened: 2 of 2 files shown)


KendallWeihe (approved):
I like this, but I wonder... should we add assertions to the deserialized variables (`deserialized_value` and `deserialized_with_type`)? The test reads from the same JSON file and then serializes it back and makes assertions after the serialization, but might there be a situation where the test passes as it is, but actually something was wrong on the deserialization of both variables? Idk if such a situation would arise

andresuribe87:
> I like this, but I wonder... should we add assertions to the deserialized variables (`deserialized_value` and `deserialized_with_type`)? The test reads from the same JSON file and then serializes it back and makes assertions after the serialization, but might there be a situation where the test passes as it is, but actually something was wrong on the deserialization of both variables? Idk if such a situation would arise

Excellent question @KendallWeihe ! It was something that I was thinking about, and decided that it's not worth it. If the situation you described happens,  it would imply that there's an underlying bug with the `serde_json::Value::Deserialize` implementation AND that its the same bug as with `PresentationDefinition::Deserialize` using the `serde_canonical_json::CanonicalFormatter`. I figured that the chance of that happening incredibly slim. Does that make sense?

For context, the original reason I wrote this test, was because there was a property in the JSON file which was valid, but there wasn't a field in the struct that represented the presentationdefinition in the typed language (or because we had a typo in a named field... can't remember).

KendallWeihe:
> there was a property in the JSON file which was valid, but there wasn't a field in the struct that represented the presentationdefinition in the typed language (or because we had a typo in a named field... can't remember)

This is 👍 and I assume this test will catch such a situation, yeah?

If we invert the situation, let's say there's a new field in the struct, and let's say it was required, but the JSON didn't have the field defined, then this test should also fail, right?

So yeah I think this is pragmatic as it's less source code overhead which covers a broader array of situations. There is a tradeoff in that it's less explicit than the prior source code, but yeah I'm cool with it ✅

andresuribe87:
> This is 👍 and I assume this test will catch such a situation, yeah?

yup!

> If we invert the situation, let's say there's a new field in the struct, and let's say it was required, but the JSON didn't have the field defined, then this test should also fail, right?

That's right, you got it!

PR #20, opened 2023-11-30
Create and resolve did:jwk

amika-sq:
This is the first DID implementation within the `crates/dids` crate, so the PR is a little bit on the larger side. I promise the next DID implementations will be MUCH smaller 😅 

The actual creation/resolution of the `did:jwk` itself is being handled by Spruce's [did-jwk](https://github.com/spruceid/ssi/tree/main/did-jwk) crate.

Beyond just create/resolve, this PR sets up the following structure:
* Define traits for `Did` and `DidMethod`. This will allow us to implement other DID methods in the future.
* Defines our own unique `DidResolutionResult`. Our upcoming test vectors expect a resolution result in a specific format. Spruce's resolution result doesn't match that format, so I defined our own that does.

Resolves https://github.com/TBD54566975/web5-rs/issues/13

### DIFF
--- a/crates/credentials/Cargo.toml
+++ b/crates/credentials/Cargo.toml
@@ -8,9 +8,9 @@ license-file.workspace = true
 
 [dependencies]
 jsonschema = "0.17.1"
-serde = { version = "1.0.193", features = ["derive"] }
-serde_json = "1.0.108"
-serde_with = "3.4.0"
+serde = { workspace = true }
+serde_json = { workspace = true }
+serde_with = { workspace = true }
 
 [dev-dependencies]
 serde_canonical_json = "1.0.0"
--- a/crates/dids/src/did.rs
+++ b/crates/dids/src/did.rs
@@ -1,0 +1,15 @@
+use crate::resolver::{DidResolutionResult, DidResolver};
+use async_trait::async_trait;
+use crypto::key_manager::KeyManager;
+use std::sync::Arc;
+
+/// Trait that defines all common behavior for a DID.
+#[async_trait]
+pub trait Did {
+    fn uri(&self) -> &str;
+    fn key_manager(&self) -> &Arc<dyn KeyManager>;
+
+    async fn resolve(&self) -> DidResolutionResult {
+        DidResolver::resolve_uri(self.uri()).await
+    }
+}
--- a/crates/dids/src/did.rs
+++ b/crates/dids/src/did.rs
@@ -6,9 +6,15 @@ use std::sync::Arc;
 /// Trait that defines all common behavior for a DID.
 #[async_trait]
 pub trait Did {
+    /// Returns the DID URI the target [`Did`] represents (e.g: `did:jwk:12345`).
     fn uri(&self) -> &str;
+
+    /// Returns a reference to the [`KeyManager`] that contains all the keys necessary to
+    /// manage and sign using the target [`Did`].
     fn key_manager(&self) -> &Arc<dyn KeyManager>;
 
+    /// Returns a [`DidResolutionResult`] for the target [`Did`], as specified in
+    /// [Resolving a DID](https://w3c-ccg.github.io/did-resolution/#resolving).
     async fn resolve(&self) -> DidResolutionResult {
         DidResolver::resolve_uri(self.uri()).await
     }
--- a/crates/dids/src/method/jwk.rs
+++ b/crates/dids/src/method/jwk.rs
@@ -1,0 +1,120 @@
+use crate::did::Did;
+use crate::method::{DidMethod, DidMethodError};
+use crate::resolver::DidResolutionResult;
+use async_trait::async_trait;
+use crypto::key::{Key, KeyType};
+use crypto::key_manager::KeyManager;
+use did_jwk::DIDJWK as SpruceDidJwkMethod;
+use ssi_dids::did_resolve::{DIDResolver, ResolutionInputMetadata};
+use ssi_dids::{DIDMethod, Source};
+use std::sync::Arc;
+
+/// Concrete implementation for a did:jwk DID
+pub struct DidJwk {
+    uri: String,
+    key_manager: Arc<dyn KeyManager>,
+}
+
+impl Did for DidJwk {
+    fn uri(&self) -> &str {
+        &self.uri
+    }
+
+    fn key_manager(&self) -> &Arc<dyn KeyManager> {
+        &self.key_manager
+    }
+}
+
+/// Options that can be used to create a did:jwk DID
+pub struct DidJwkCreateOptions {
+    pub key_type: KeyType,
+}
+
+#[async_trait]
+impl DidMethod<DidJwk, DidJwkCreateOptions> for DidJwk {
+    const NAME: &'static str = "jwk";
+
+    fn create(
+        key_manager: Arc<dyn KeyManager>,
+        options: DidJwkCreateOptions,
+    ) -> Result<DidJwk, DidMethodError> {
+        let key_alias = key_manager.generate_private_key(options.key_type)?;
+        let public_key =
+            key_manager
+                .get_public_key(&key_alias)?
+                .ok_or(DidMethodError::DidCreationFailure(
+                    "PublicKey not found".to_string(),
+                ))?;
+
+        let uri = SpruceDidJwkMethod
+            .generate(&Source::Key(&public_key.jwk()))
+            .ok_or(DidMethodError::DidCreationFailure(
+                "Failed to generate did:jwk".to_string(),
+            ))?;
+
+        Ok(DidJwk { uri, key_manager })
+    }
+
+    async fn resolve_uri(did_uri: &str) -> DidResolutionResult {
+        let input_metadata = ResolutionInputMetadata::default();
+        let (did_resolution_metadata, did_document, did_document_metadata) =
+            SpruceDidJwkMethod.resolve(did_uri, &input_metadata).await;
+
+        DidResolutionResult {
+            did_resolution_metadata,
+            did_document,
+            did_document_metadata,
+            ..Default::default()
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use ssi_dids::did_resolve::ERROR_INVALID_DID;
+
+    fn create_did_jwk() -> DidJwk {
+        let key_manager = Arc::new(crypto::key_manager::LocalKeyManager::new_in_memory());
+        let options = DidJwkCreateOptions {
+            key_type: KeyType::Ed25519,
+        };
+
+        DidJwk::create(key_manager, options).unwrap()
+    }
+
+    #[test]
+    fn create_produces_correct_uri() {
+        let did = create_did_jwk();
+        assert!(did.uri.starts_with("did:jwk:"));
+    }
+
+    #[tokio::test]
+    async fn instance_resolve() {
+        let did = create_did_jwk();
+        let result = did.resolve().await;
+        assert!(result.did_resolution_metadata.error.is_none());
+
+        let did_document = result.did_document.unwrap();
+        assert_eq!(did_document.id, did.uri);
+    }
+
+    #[tokio::test]
+    async fn resolve_uri_success() {
+        let did = create_did_jwk();
+        let result = DidJwk::resolve_uri(&did.uri).await;
+        assert!(result.did_resolution_metadata.error.is_none());
+
+        let did_document = result.did_document.unwrap();
+        assert_eq!(did_document.id, did.uri);
+    }
+
+    #[tokio::test]
+    async fn resolve_uri_failure() {
+        let result = DidJwk::resolve_uri("did:jwk:does-not-exist").await;
+        assert_eq!(
+            result.did_resolution_metadata.error,
+            Some(ERROR_INVALID_DID.to_string())
+        );
+    }
+}
--- a/crates/dids/src/method/mod.rs
+++ b/crates/dids/src/method/mod.rs
@@ -18,6 +18,15 @@ pub enum DidMethodError {
 /// A trait with common behavior across all DID methods.
 #[async_trait]
 pub trait DidMethod<T: Did, CreateOptions> {
+    /// The name of the implemented DID method (e.g. `jwk`).
+    ///
+    /// This is used to identify the [`DidMethod`] responsible for creating/resolving an arbitrary
+    /// DID URI.
+    ///
+    /// # Example
+    /// If a consumer wants to resolve a DID URI of `did:jwk:12345`, the method portion of the URI
+    /// (`jwk` in this example) is compared against each [`DidMethod`]'s `NAME` constant. If a match
+    /// is found, the corresponding [`DidMethod`] is used to resolve the DID URI.
     const NAME: &'static str;
 
     /// Create a new DID instance.
@@ -26,6 +35,7 @@ pub trait DidMethod<T: Did, CreateOptions> {
         options: CreateOptions,
     ) -> Result<T, DidMethodError>;
 
-    /// Resolve a DID URI to a DID Document.
+    /// Resolve a DID URI to a [`DidResolutionResult`], as specified in
+    /// [Resolving a DID](https://w3c-ccg.github.io/did-resolution/#resolving).
     async fn resolve_uri(did_uri: &str) -> DidResolutionResult;
 }
--- a/crates/dids/src/method/mod.rs
+++ b/crates/dids/src/method/mod.rs
@@ -1,0 +1,31 @@
+pub mod jwk;
+
+use crate::did::Did;
+use crate::resolver::DidResolutionResult;
+use async_trait::async_trait;
+use crypto::key_manager::{KeyManager, KeyManagerError};
+use std::sync::Arc;
+
+/// Errors that can occur when working with DID methods.
+#[derive(thiserror::Error, Debug)]
+pub enum DidMethodError {
+    #[error(transparent)]
+    KeyManagerError(#[from] KeyManagerError),
+    #[error("Failure creating DID: {0}")]
+    DidCreationFailure(String),
+}
+
+/// A trait with common behavior across all DID methods.
+#[async_trait]
+pub trait DidMethod<T: Did, CreateOptions> {
+    const NAME: &'static str;
+
+    /// Create a new DID instance.
+    fn create(
+        key_manager: Arc<dyn KeyManager>,
+        options: CreateOptions,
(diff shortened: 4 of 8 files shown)


crates/dids/src/resolver.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    #[tokio::test]
+    async fn resolve_did_jwk() {
+        let did_uri = "did:jwk:eyJjcnYiOiJQLTI1NiIsImt0eSI6IkVDIiwieCI6ImFjYklRaXVNczNpOF91c3pFakoydHBUdFJNNEVVM3l6OTFQSDZDZEgyVjAiLCJ5IjoiX0tjeUxqOXZXTXB0bm1LdG00NkdxRHo4d2Y3NEk1TEtncmwyR3pIM25TRSJ9";
+        let result = DidResolver::resolve_uri(did_uri).await;
+        assert!(result.did_resolution_metadata.error.is_none());
+
+        let did_document = result.did_document.unwrap();
+        assert_eq!(did_document.id, did_uri);
+    }
```
amika-sq, line 88: I currently do not consume the test vector in this PR, as I wanted to keep it as small as I could. Going to be adding that in a follow-up PR to this. For now, I resolve the `did:jwk` as defined from the first vector [here](https://tbd54566975.github.io/sdk-development/web5-test-vectors/did-jwk/resolve.json).

crates/credentials/Cargo.toml
```diff
@@ -8,9 +8,9 @@ license-file.workspace = true
 
 [dependencies]
 jsonschema = "0.17.1"
-serde = { version = "1.0.193", features = ["derive"] }
-serde_json = "1.0.108"
-serde_with = "3.4.0"
+serde = { workspace = true }
+serde_json = { workspace = true }
+serde_with = { workspace = true }
```
amika-sq, line 13: the `credentials` and `dids` crate both now depend on serde crates. Moved them to the workspace `Cargo.toml` file so that their versions are always in sync with each other.

crates/dids/src/method/jwk.rs
```diff
@@ -0,0 +1,120 @@
...
+        let uri = SpruceDidJwkMethod
+            .generate(&Source::Key(&public_key.jwk()))
+            .ok_or(DidMethodError::DidCreationFailure(
+                "Failed to generate did:jwk".to_string(),
+            ))?;
+
+        Ok(DidJwk { uri, key_manager })
+    }
+
+    async fn resolve_uri(did_uri: &str) -> DidResolutionResult {
```
amika-sq, line 58: There's a lot of code within TBD that use the term "did." Depending on where you're looking, it could mean one of two things:
1. The unique DID string
2. A structured representation of a DID

In this PR, I chose to always refer to the unique DID string as `did_uri` to hopefully help clarity while reading the code. 

Looking for feedback on this! 
Do you love it? Do you hate it? Happy to change if there's any strong feelings here.

crates/dids/src/method/jwk.rs
```diff
@@ -0,0 +1,120 @@
...
+        DidJwk::create(key_manager, options).unwrap()
+    }
+
+    #[test]
+    fn create_produces_correct_uri() {
+        let did = create_did_jwk();
+        assert!(did.uri.starts_with("did:jwk:"));
+    }
+
+    #[tokio::test]
```
KendallWeihe, line 92: So why `tokio`?

crates/dids/src/resolver.rs
```diff
@@ -0,0 +1,109 @@
...
+use ssi_dids::did_resolve::{
+    DocumentMetadata as DidDocumentMetadata, ResolutionMetadata as DidResolutionMetadata,
+};
+use ssi_dids::Document as DidDocument;
+
+pub struct DidResolver;
+
+impl DidResolver {
+    /// Resolves a DID URI, using the appropriate DID method, to a DID Document.
+    pub async fn resolve_uri(did_uri: &str) -> DidResolutionResult {
```
KendallWeihe, line 13: So the idea here being, the DX is such that they can have a generalized `DidResolver.resolve_uri()` method to which they can pass in any URI and so long as it's one of our supported methods, and then get the resolution result back? 

This would be in contrast to a DX where the developer is already aware of their chosen DID Method, and they directly call the method resolution, for example `DidJwk.resolve_uri()`

I think that's a fine DX... okay and I can see we're offering [the same DX in `web5-js`](https://github.com/TBD54566975/web5-js/blob/main/packages/dids/src/did-resolver.ts#L63). So alright cool, just thinking out loud ✅

crates/dids/src/method/jwk.rs
KendallWeihe: I presume it has something to do with us actually testing an external network integration

crates/dids/src/resolver.rs
KendallWeihe: cool, thanks for keeping things concise here

KendallWeihe (approved):
nice! lgtm!

crates/dids/src/method/jwk.rs
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    fn key_manager(&self) -> &Arc<dyn KeyManager> {
+        &self.key_manager
+    }
+}
+
+/// Options that can be used to create a did:jwk DID
+pub struct DidJwkCreateOptions {
+    pub key_type: KeyType,
```
andresuribe87, line 30: I actually found `KeyType` to be a bit confusing. Looking at the docs for `KeyType` didn't actually help. 

Is `KeyType` meant to represent https://www.rfc-editor.org/rfc/rfc7518.html#section-6.1 ? Or is it meant to represent the name of the [elliptic curve](https://www.iana.org/assignments/jose/jose.xhtml#web-key-elliptic-curve)?

In kotlin, we defined the creation options as Algo and Curve, curious how/why this diverged?

crates/dids/src/method/jwk.rs
andresuribe87: Makes sense, it's where we landed on the KT side (in most places).

crates/dids/src/method/jwk.rs
andresuribe87: DidJwk resolution should not go over the network. Chad tells me that

> The #[tokio::test] attribute is specific to the tokio runtime and is used to mark an asynchronous test function. It indicates that the following function is an asynchronous test that should be executed using the tokio runtime.

crates/dids/src/resolver.rs
andresuribe87: Similar stuff is done in kt and golang. While this is fine, my main feedback is that this mechanism does not allow consumers to configure `DidResolver` to use consumer specified http_client libraries, nor which gateways to use for resolution.

crates/dids/src/method/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+    #[error(transparent)]
+    KeyManagerError(#[from] KeyManagerError),
+    #[error("Failure creating DID: {0}")]
+    DidCreationFailure(String),
+}
+
+/// A trait with common behavior across all DID methods.
+#[async_trait]
+pub trait DidMethod<T: Did, CreateOptions> {
+    const NAME: &'static str;
```
andresuribe87, line 21: What do you think about adding documentation that tells implementers what they should set this constant to, and how this is helpful? I.e. "jwk" or "key" or whatever.

crates/dids/src/did.rs
```diff
@@ -0,0 +1,15 @@
+use crate::resolver::{DidResolutionResult, DidResolver};
+use async_trait::async_trait;
+use crypto::key_manager::KeyManager;
+use std::sync::Arc;
+
+/// Trait that defines all common behavior for a DID.
+#[async_trait]
+pub trait Did {
+    fn uri(&self) -> &str;
```
andresuribe87, line 9:
```suggestion
    /// Returns the did_uri that this Did structure represents. E.g. `did:jwk:foo`. 
    fn uri(&self) -> &str;
```

crates/dids/src/did.rs
```diff
@@ -0,0 +1,15 @@
...
+    fn key_manager(&self) -> &Arc<dyn KeyManager>;
```
andresuribe87, line 10:
```suggestion
    /// Returns a reference to a KeyManager implementation that contains all the keys necessary to manage and sign using this Did.
    fn key_manager(&self) -> &Arc<dyn KeyManager>;
```

crates/dids/src/did.rs
```diff
@@ -0,0 +1,15 @@
...
+
+    async fn resolve(&self) -> DidResolutionResult {
```
andresuribe87, line 12: I think? 
```suggestion
    /// Returns a DidResolutionResult as specified in https://www.w3.org/TR/did-core/#did-resolution
    async fn resolve(&self) -> DidResolutionResult {
```

crates/dids/src/method/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+
+    /// Create a new DID instance.
+    fn create(
+        key_manager: Arc<dyn KeyManager>,
+        options: CreateOptions,
+    ) -> Result<T, DidMethodError>;
+
+    /// Resolve a DID URI to a DID Document.
```
andresuribe87, line 29:
```suggestion
    /// Resolve a DID URI to a DidResolutionResult.
```

crates/dids/src/resolver.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    /// Returns the method name of a DID URI, if the provided DID URI is valid, `None` otherwise.
+    fn method_name(did_uri: &str) -> Option<&str> {
+        let parts: Vec<&str> = did_uri.split(':').collect();
+        if parts.len() < 3 || parts[0] != "did" {
+            return None;
+        };
+
+        Some(parts[1])
+    }
```
andresuribe87, line 33: I would expect this code to be inside the `Did` trait. Thoughts?

crates/dids/src/method/jwk.rs
```diff
@@ -0,0 +1,120 @@
...
+use did_jwk::DIDJWK as SpruceDidJwkMethod;
+use ssi_dids::did_resolve::{DIDResolver, ResolutionInputMetadata};
+use ssi_dids::{DIDMethod, Source};
+use std::sync::Arc;
+
+/// Concrete implementation for a did:jwk DID
+pub struct DidJwk {
+    uri: String,
+    key_manager: Arc<dyn KeyManager>,
+}
```
andresuribe87, line 16: When creating an instance of `DidJwk`, we're going to add validation in kt. This validation will check that all the required keys for managing the given `did_uri` already exist within the `key_manager`. The bulk of this logic exists in https://github.com/TBD54566975/web5-kt/blob/dbc1df5aba9206a35791e6a20f6aa26acddc8e7f/dids/src/main/kotlin/web5/sdk/dids/DidMethod.kt#L193C1-L193C1

Would it make sense to do something similar here? No idea how you could do so in rust.

andresuribe87 (commented):
This is awesome, some question/suggestion from what we've seen in Kt.

crates/dids/src/method/jwk.rs
amika-sq: @andresuribe87 has it right. 

Resolution is generic across all DID methods, and resolution CAN be an asynchronous operation for certain methods. So the `DidMethod` trait defines the resolve function as asynchronous.

While `did:jwk` does not resolve over the network, the function must be asynchronous to conform to the `DidMethod` trait. 

We have a few options to test asynchronous functions:
1. Write a normal, synchronous, test. Whenever you need to call an asynchronous function within the test, block the thread until the asynchronous function completes.
2. Use `tokio` or `async-std` (amongst others). 

Tokio is much more widely adopted, so I went with that. Happy to alter though if there's any strong opinions another way!

crates/dids/src/method/jwk.rs
amika-sq: I originally had it as `KeyAlgorithm`, but was recommended to rename it to `KeyType` here: https://github.com/TBD54566975/web5-rs/pull/17#discussion_r1401124839

Each of the algorithms use the same curve in Kotlin, so I just didn't add them to the Rust version yet. Can add them later if/when we use them.

crates/dids/src/method/mod.rs
amika-sq: I'll update and submit a PR to the Kotlin side as well here.

crates/dids/src/resolver.rs
amika-sq: The `Did` trait defines behavior for `Did` implementations. Each implementation is already defining the name it's implementing. This function is computing the method name from some arbitrarily provided string.

We could definitely move this into a `utils` module or something similar.

crates/dids/src/method/mod.rs
amika-sq: I'll add a little blurb here.

I'm going to be adding documentation around how to implement additional DID methods once this, did:key, and did:web land. Was thinking it'll be a README within this `method` module.

crates/dids/src/resolver.rs
amika-sq: Gateways are really did-method specific, and I don't necessarily think that it's needed in this general "resolve this did_uri string for me" function. 

If a caller wants to resolve with a particular gateway, they probably already know which DidMethod they want to resolve with. They could directly call that resolver, with whatever options they want.

I'm a bit curious though on the http_client libraries. What did you have in mind there?

crates/dids/src/did.rs
KendallWeihe: agreed w/ @amika-sq

crates/dids/src/resolver.rs
KendallWeihe: The answer is never "utils"

crates/dids/src/method/jwk.rs
```diff
@@ -0,0 +1,120 @@
...
+                .get_public_key(&key_alias)?
+                .ok_or(DidMethodError::DidCreationFailure(
+                    "PublicKey not found".to_string(),
+                ))?;
+
+        let uri = SpruceDidJwkMethod
+            .generate(&Source::Key(&public_key.jwk()))
+            .ok_or(DidMethodError::DidCreationFailure(
+                "Failed to generate did:jwk".to_string(),
+            ))?;
```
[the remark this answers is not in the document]
decentralgabe: One thing to note - the did:jwk spec is being moved to the IETF with some changes now (and more coming later). We will need to be agile with our implementation, so having more control is a benefit. For example, we'll soon need to add in optional did:dht resolution for jwk.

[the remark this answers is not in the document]
amika-sq: We can totally add any changes if/when needed! But right now, there's no difference. I'd be copy/pasting Spruce's implementation. If that's what we want, I can do it!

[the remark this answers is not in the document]
andresuribe87: It also puts us in a better position for supply chain security by minimizing deps.

crates/dids/src/resolver.rs
amika-sq: @andresuribe87 What would you suggest here then? Should we modify this PR to add in options to resolve with a gateway? `did:jwk` doesn't take a gateway or do anything over http, so it would just be for future functionality right now. 

I don't mind the suggestion, but I don't know how we'd actionably do something with it here yet.

crates/dids/src/method/jwk.rs
andresuribe87: `DidJwk` is a public struct. IIUC, consumers can instantiate is by doing something like the example below. 

```rs
let my_did_jwk = DidJwk {
    uri: String::from("did:jwk:example"),
    key_manager: Arc::clone(keyManager), // keyManager does not have the keys for `did:jwk:example`!!
};
```
Is my understanding correct? 

If so, then my suggestion is to either: a) prevent consumers from doing such a construction (since you cannot use `my_did_jwk` to sign stuff if it doesn't have the keys) or b) validate that the keys exist inside the `keyManager` upon instantiation. Is something like that reasonable?

crates/dids/src/method/jwk.rs
decentralgabe: yes I was referring to the concept similar to `kty` https://www.rfc-editor.org/rfc/rfc7518.html#section-6.1

crates/dids/src/method/jwk.rs
amika-sq: Your understanding isn't quite correct here. That manner of construction is only supported _within_ the `dids` crate due to Rust's access control.

Here's an example of someone trying to consume the `dids` crate and create a new `DidJwk` in the manner you suggest:

Screenshot 2023-12-04 at 11 30 32 AM

This will not compile, as `uri` and `key_manager` are not public. It will result in the following compilation error:
```
❯ cargo build -p example
   Compiling example v0.1.0 (/Users/amika/Development/rust/web5-rs/crates/show-andres-hes-wrong)
error[E0451]: field `uri` of struct `DidJwk` is private
 --> crates/show-andres-hes-wrong/src/lib.rs:7:9
  |
7 |         uri: "did:jwk:1234".to_string(),
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ private field

error[E0451]: field `key_manager` of struct `DidJwk` is private
 --> crates/show-andres-hes-wrong/src/lib.rs:8:9
  |
8 |         key_manager: Arc::new(LocalKeyManager::new_in_memory()),
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ private field

For more information about this error, try `rustc --explain E0451`.
error: could not compile `example` (lib) due to 2 previous errors
```

If we want external users of the `dids` crate to create arbitrary `DidJwk` objects, we would either need to do one of the following:
1. Make `uri` and `key_manager` public on `DidJwk`
2. Make a public constructor that allows the user to provide their own `uri` and `key_manager`

I would prefer option 2, as we can encapsulate validation logic within a constructor.

andresuribe87 (approved):
None of my comments are blocking, but I think there's good conversations that we need to figure out.

crates/dids/src/method/jwk.rs
andresuribe87: Ohhh thank you so much. That makes a ton of sense and I appreciate the explanation!

crates/dids/src/method/jwk.rs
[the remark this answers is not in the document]
mistermoe: unless we have concerns about how `did:jwk` is currently implemented in spruce's sdk, i don't see why we'd implement it ourselves at the moment

PR #24, opened 2023-12-01
Resolve did:web

amika-sq:
Resolve `did:web` DIDs using Spruce's [did-web](https://crates.io/crates/did-web) crate. Uses our `DidResolutionResult` implementation so that the resolved DIDDocument & all respective metadata is in the expected format according to the community draft resolution spec [here](https://w3c-ccg.github.io/did-resolution/#resolving).

Resolves https://github.com/TBD54566975/web5-rs/issues/12

### DIFF
--- a/crates/dids/src/method/web.rs
+++ b/crates/dids/src/method/web.rs
@@ -1,0 +1,82 @@
+use crate::did::Did;
+use crate::method::{DidMethod, DidMethodError, DidResolutionResult};
+use async_trait::async_trait;
+use crypto::key_manager::KeyManager;
+use did_web::DIDWeb as SpruceDidWebMethod;
+use ssi_dids::did_resolve::{DIDResolver, ResolutionInputMetadata};
+use std::sync::Arc;
+
+/// Concrete implementation for a did:web DID
+pub struct DidWeb {
+    uri: String,
+    key_manager: Arc<dyn KeyManager>,
+}
+
+impl Did for DidWeb {
+    fn uri(&self) -> &str {
+        &self.uri
+    }
+
+    fn key_manager(&self) -> &Arc<dyn KeyManager> {
+        &self.key_manager
+    }
+}
+
+/// Options that can be used to create a did:web DID.
+/// This is currently a unit struct because did:web does not support key creation.
+pub struct DidWebCreateOptions;
+
+#[async_trait]
+impl DidMethod<DidWeb, DidWebCreateOptions> for DidWeb {
+    const NAME: &'static str = "web";
+
+    fn create(
+        _key_manager: Arc<dyn KeyManager>,
+        _options: DidWebCreateOptions,
+    ) -> Result<DidWeb, DidMethodError> {
+        Err(DidMethodError::DidCreationFailure(
+            "create operation not supported for did:web".to_string(),
+        ))
+    }
+
+    async fn resolve_uri(did_uri: &str) -> DidResolutionResult {
+        let input_metadata = ResolutionInputMetadata::default();
+        let (did_resolution_metadata, did_document, did_document_metadata) =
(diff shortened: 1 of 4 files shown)


crates/dids/src/method/web.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    #[test]
+    fn create_fails() {
+        let key_manager = Arc::new(crypto::key_manager::LocalKeyManager::new_in_memory());
+        let result = DidWeb::create(key_manager, DidWebCreateOptions);
+        assert!(result.is_err());
+    }
+
+    #[tokio::test]
+    async fn resolution_success() {
```
decentralgabe, line 68: is this doing a real resolution? should it be mocked?

amika-sq: Yes, it's a real resolution. I can add a mock if we want to.

amika-sq: For reference, Spruce is testing mocked responses here: https://github.com/spruceid/ssi/blob/main/did-web/src/lib.rs#L197-L218

decentralgabe: I actually prefer not mocking but I believe we're mocking elsewhere in our libs. Just flagging because the test could fail even if it's correct, since it relies on an external resource. if there's a way to distinguish network tests somehow that would be useful.

crates/dids/src/method/web.rs
```diff
@@ -0,0 +1,82 @@
...
+        let did_uri = "did:web:tbd.website";
```
andresuribe87, line 69: Would be nice to add a test like `did:web:users.tbddev.org:demo` which should resolve to https://users.tbddev.org/demo/did.json

andresuribe87 (approved):
Awesome work!

KendallWeihe (approved):
nice this lgtm, but I think we should make the change @andresuribe87 suggested around the non-empty resolution

crates/dids/src/method/web.rs
amika-sq: Is this valid? I'm trying to use https://dev.uniresolver.io/ and it's not able to resolve this 🤔

amika-sq: Did a little digging. Resolving this DID currently fails because it's using an unknown VerificationMethod property `publicKeyHex`. This doesn't appear in the spec: https://www.w3.org/TR/did-core/#verification-method-properties

Is there another spec that does include this? If so, why have these diverged?

@decentralgabe maybe you have some extra insight here?

decentralgabe: @amika-sq it's not included in the default context for did-core, but it is included in the security vocab here -- https://github.com/w3c-ccg/security-vocab/pull/40/files

this means to make that did web valid it would have to add the context like this

```json
{
  "@context": ["https://w3id.org/did/v1", "https://w3id.org/security/v3-unstable"],
  "id": "did:web:users.tbddev.org:demo",
  "verificationMethod": [
    {
      "id": "did:web:users.tbddev.org:demo#key-0",
      "type": "EcdsaSecp256k1VerificationKey2019",
      "publicKeyHex": "04e8e82698e64db79e13dc6e7478f47a5fe867c4acd6fa8928b0c93c763a02fd6c54a622de247cc04449531429dbb6a2373b6c685a8e1326913e0531e5f649208d"
    }
  ]
}
```

or not use hex

decentralgabe: really we should be using JWKs everywhere.

amika-sq: Makes sense! I went ahead and added a task in the backlog for this here: https://github.com/TBD54566975/web5-rs/issues/39

I'm thinking we can add one specifically for testing purposes [here](https://github.com/TBD54566975/web5-rs/issues/39) so that we have a reliable way to always resolve.

Issue #22, opened 2023-11-30
Implement `did:key` creation and resolution


amika-sq:
We can consume Spruce's implementation [did-method-key](https://github.com/spruceid/ssi/tree/main/did-key) for the brains of this operation.

PR #25, opened 2023-12-01
Create and resolve did:key

amika-sq:
Resolve `did:key` DIDs using Spruce's [did-method-key](https://crates.io/crates/did-method-key) crate. Uses our DidResolutionResult implementation so that the resolved DIDDocument & all respective metadata is in the expected format according to the community draft resolution spec [here](https://w3c-ccg.github.io/did-resolution/#resolving).

Resolves https://github.com/TBD54566975/web5-rs/issues/22

### DIFF
--- a/crates/dids/src/method/key.rs
+++ b/crates/dids/src/method/key.rs
@@ -73,24 +73,29 @@ mod tests {
     use crypto::key_manager::local_key_manager::LocalKeyManager;
     use ssi_dids::did_resolve::ERROR_INVALID_DID;
 
-    fn create_did_key() -> DidKey {
+    fn create_did_key(key_type: KeyType) -> DidKey {
         let key_manager = Arc::new(LocalKeyManager::new_in_memory());
-        let options = DidKeyCreateOptions {
-            key_type: KeyType::Ed25519,
-        };
 
-        DidKey::create(key_manager, options).expect("DidKey creation failed")
+        DidKey::create(key_manager, DidKeyCreateOptions { key_type })
+            .expect("DidKey creation failed")
     }
 
     #[test]
     fn create_produces_correct_uri() {
-        let did = create_did_key();
+        let did = create_did_key(KeyType::Ed25519);
         assert!(did.uri.starts_with("did:key:"));
+    }
+
+    #[test]
+    fn create_each_key_type() {
+        create_did_key(KeyType::Ed25519);
+        create_did_key(KeyType::Secp256k1);
+        create_did_key(KeyType::Secp256r1);
     }
 
     #[tokio::test]
     async fn instance_resolve() {
-        let did = create_did_key();
+        let did = create_did_key(KeyType::Ed25519);
         let result = did.resolve().await;
         assert!(result.did_resolution_metadata.error.is_none());
 
@@ -100,7 +105,7 @@ mod tests {
 
     #[tokio::test]
     async fn resolve_uri_success() {
-        let did = create_did_key();
+        let did = create_did_key(KeyType::Ed25519);
         let result = DidKey::resolve_uri(&did.uri).await;
         assert!(result.did_resolution_metadata.error.is_none());
 
--- a/crates/dids/src/method/key.rs
+++ b/crates/dids/src/method/key.rs
@@ -1,0 +1,119 @@
+use crate::did::Did;
+use crate::method::{DidMethod, DidMethodError, DidResolutionResult};
+use async_trait::async_trait;
+use crypto::key::{Key, KeyType};
+use crypto::key_manager::KeyManager;
+use did_method_key::DIDKey as SpruceDidKeyMethod;
+use ssi_dids::did_resolve::{DIDResolver, ResolutionInputMetadata};
+use ssi_dids::{DIDMethod, Source};
+use std::sync::Arc;
+
+/// Concrete implementation for a did:key DID
+pub struct DidKey {
+    uri: String,
+    key_manager: Arc<dyn KeyManager>,
+}
+
+pub struct DidKeyCreateOptions {
+    pub key_type: KeyType,
+}
+
+impl Did for DidKey {
+    fn uri(&self) -> &str {
+        &self.uri
+    }
+
+    fn key_manager(&self) -> &Arc<dyn KeyManager> {
+        &self.key_manager
+    }
+}
+
+#[async_trait]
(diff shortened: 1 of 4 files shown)


amika-sq:
Created a task in our backlog to implement this ourselves: https://github.com/TBD54566975/web5-rs/issues/43

crates/dids/src/method/key.rs
```diff
@@ -0,0 +1,119 @@
...
+        assert_eq!(did_document.id, did.uri);
+    }
+
+    #[tokio::test]
+    async fn resolve_uri_success() {
+        let did = create_did_key();
+        let result = DidKey::resolve_uri(&did.uri).await;
+        assert!(result.did_resolution_metadata.error.is_none());
+
+        let did_document = result.did_document.expect("did_document not found");
```
KendallWeihe, line 107: I'm confused, why would the resolution yield a DID Document "not found" result?

crates/dids/src/method/key.rs
```diff
@@ -0,0 +1,119 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crypto::key_manager::local_key_manager::LocalKeyManager;
+    use ssi_dids::did_resolve::ERROR_INVALID_DID;
+
+    fn create_did_key() -> DidKey {
+        let key_manager = Arc::new(LocalKeyManager::new_in_memory());
+        let options = DidKeyCreateOptions {
+            key_type: KeyType::Ed25519,
```
decentralgabe, line 79: would recommend adding tests for the key types we care about

crates/dids/src/method/key.rs
amika-sq: `expect` is a function on `Result`: https://doc.rust-lang.org/std/result/enum.Result.html#method.expect

This line is essentially saying:
`result.did_document` is expected to be in the `Ok` part of the result. If it doesn't, panic with the message "did_document not found".

KendallWeihe: Oooooh, cool

crates/dids/src/method/key.rs
```diff
@@ -0,0 +1,124 @@
...
+mod tests {
+    use super::*;
+    use crypto::key_manager::local_key_manager::LocalKeyManager;
+    use ssi_dids::did_resolve::ERROR_INVALID_DID;
+
+    fn create_did_key(key_type: KeyType) -> DidKey {
+        let key_manager = Arc::new(LocalKeyManager::new_in_memory());
+
+        DidKey::create(key_manager, DidKeyCreateOptions { key_type })
+            .expect("DidKey Ed25519 creation failed")
```
KendallWeihe, line 80: Error message is hardcoded to `Ed25519` as the key type rather than using the `key_type` parameter

KendallWeihe: This actually begs the question... should we define default key types?

amika-sq: Fixed it as I rebased to main!

Rust doesn't have the concept of "default parameter values" like a lot of other languages have. We _could_ define a default value for ANY `KeyType` if it's not present, but that didn't seem right to me. What may be a good default for one did method may not be a good default for another.

The other approach we could take is the builder pattern, which would allow us to define default values if the builder doesn't have a value set for a required field.

amika-sq: Another options is we could conform `DidKeyCreateOptions` (and the others like `DidJwkCreateOptions`) and have it conform to the `Default` trait ([documentation](https://doc.rust-lang.org/std/default/trait.Default.html)).

Then we could define whatever default options we think are good for creating `did:key` with.

The only downside, the user still needs to provide a value when calling this function. It would just change to the following:
`DidKey::create(key_manager, DidKeyCreateOptions::default())`

KendallWeihe: Just briefly perusing `web5-js` and `web5-kt`, we're offering at least some amount of default-value DX. I think this warrants an issue on the backlog, if only for the sake of discussion to rule it out. I don't love simplicity insofar as it robs the developer of valuable knowledge, but intuitively, I think it would be wise to offer a simple `create()` DX where the developer can easily get-off-zero. Which further begs the question of if the key manager should also have a default value.

KendallWeihe: A compelling counter argument, against supporting default values, is that Rust is not a junior dev friendly language. Which is to say, not a lot of newbie developers are using Rust. In which case, default values only serve to add unnecessary overhead and confusion. 

Idk just trying to steel man the counter case, I still think intuitively, default values are a good DX.

KendallWeihe: Late in the day for me so I'm rambling, but anyways this isn't a priority whatsoever IMO

KendallWeihe: https://github.com/TBD54566975/web5-rs/issues/45

crates/dids/src/method/key.rs
```diff
@@ -0,0 +1,124 @@
...
+    fn key_manager(&self) -> &Arc<dyn KeyManager> {
+        &self.key_manager
+    }
+}
+
+#[async_trait]
+impl DidMethod<DidKey, DidKeyCreateOptions> for DidKey {
+    const NAME: &'static str = "key";
+
+    fn create(
```
andresuribe87, line 35: Does it make sense to add documentation to public facing functions?

andresuribe87 (approved):
Small suggestion about public facing docs, so when documentation is automatically generated then readers have clarity.

crates/dids/src/method/key.rs
amika-sq: This isn't public. `DidMethod` is public, and it is documented here: https://github.com/TBD54566975/web5-rs/blob/main/crates/dids/src/method/mod.rs#L21-L42

andresuribe87: > This isn't public.

The implementation of `DidMethod::create` is being done via `DidKey::create`. Since the former is public, I thought that meant this method is public as well....

Anyways, I don't think it's important. My main concern is that there is no place in documentation that describes to consumers of the crate how to use `DidKeyCreateOptions` when creating a did of type key. While this is obvious for `DidKey`, it will become not so obvious for `DidDht` in the future.

amika-sq: We can address this when we implement `DidDht`

PR #34, opened 2023-12-04
Rename KeyType -> KeyAlgorithm

amika-sq:
`KeyType` was found to be a bit confusing: https://github.com/TBD54566975/web5-rs/pull/20#discussion_r1412220155

We originally had it as `KeyAlgorithm`, but changed it here: https://github.com/TBD54566975/web5-rs/pull/17#discussion_r1401124839

This PR switches it back to `KeyAlgorithm`!

### DIFF
--- a/crates/crypto/src/key_manager/key_manager.rs
+++ b/crates/crypto/src/key_manager/key_manager.rs
@@ -1,4 +1,4 @@
-use crate::key::{KeyError, KeyAlgorithm, PublicKey};
+use crate::key::{KeyAlgorithm, KeyError, PublicKey};
 use crate::key_manager::key_store::KeyStoreError;
 use ssi_jwk::Error as JWKError;
 
--- a/crates/crypto/src/key_manager/local_key_manager.rs
+++ b/crates/crypto/src/key_manager/local_key_manager.rs
@@ -91,7 +91,9 @@ mod tests {
     fn test_get_public_key() {
         let key_manager = LocalKeyManager::new_in_memory();
 
-        let key_alias = key_manager.generate_private_key(KeyAlgorithm::Ed25519).unwrap();
+        let key_alias = key_manager
+            .generate_private_key(KeyAlgorithm::Ed25519)
+            .unwrap();
 
         key_manager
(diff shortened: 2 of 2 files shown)


crates/crypto/src/key/key.rs
```diff
@@ -1,8 +1,8 @@
 use ssi_jwk::JWK;
 use ssi_jws::Error as JWSError;
 
-/// Enum defining all supported cryptographic key types.
-pub enum KeyType {
+/// Enum defining all supported cryptographic algorithms for a [`Key`].
+pub enum KeyAlgorithm {
```
andresuribe87, line 5: The values for this enum appear to me to be Elliptic Curves (as defined in https://www.rfc-editor.org/rfc/rfc7518.html#section-6.2.1.1 and who's possible values are in the registry located on https://www.iana.org/assignments/jose/jose.xhtml#web-key-elliptic-curve). How about we call this `Curve`?

decentralgabe: supporting RSA 😄

andresuribe87: Curious what thought are here, specially from @mistermoe and @decentralgabe

decentralgabe (changes requested):
This change is not semantically correct. We need to align with industry used terms, like key type (`kty` in JOSE land). https://datatracker.ietf.org/doc/html/rfc7517#section-4.1

Algorithms are used with _keys_ which have a _type_. Like:

| Key Type         | Signature Algorithm | 
|------------------|---------------------|
| Ed25519          | EdDSA               |
| secp256k1        | ES256K              |
| P-256            | ES256               |
| P-384            | ES384               |
| P-521            | ES512               |
| RSA              | PS256               |
| BLS              | BBS+                |
| Dilithium Mode 2 | CRYDI2              |
| Dilithium Mode 3 | CRYDI3              |
| Dilithium Mode 5 | CRYDI5              |

amika-sq:
Point taken, closing!

KendallWeihe:
Thanks for sharing @decentralgabe, I wasn't aware of the distinction

PR #35, opened 2023-12-04
Add linter to CI and applied some fixes

andresuribe87:
This PR adds the clippy linter to the repo. See https://doc.rust-lang.org/clippy/index.html for more information. 

I applied the automatic fixes. There's still one more error: https://rust-lang.github.io/rust-clippy/master/index.html#/module_inception, which we need to decide how to handle. Open for feedback on how to best deal with it.


.github/workflows/ci.yml
```diff
@@ -10,14 +10,25 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
+  # Make sure CI fails on all warnings, including Clippy lints
+  RUSTFLAGS: "-Dwarnings"
 
 jobs:
+  clippy_check:
```
amika-sq, line 17:
```suggestion
  clippy:
```

Just `clippy` is fine.

.github/workflows/ci.yml
```diff
@@ -10,14 +10,25 @@ on:
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust: [ stable, nightly ]
+    steps:
+      - uses: actions/checkout@v3
+      - name: Run Clippy
+        run: cargo clippy --all-targets --all-features
```
amika-sq, line 25:
```suggestion
        run: cargo clippy --workspace
```

With the way our workspace is setup currently, we'll need to use the workspace flag instead.

amika-sq (commented):
Looks good! Few minor comments:

> I applied the automatic fixes. There's still one more error: https://rust-lang.github.io/rust-clippy/master/index.html#/module_inception, which we need to decide how to handle. Open for feedback on how to best deal with it.

I can take this one on and fix it. We consolidated on the module naming structure here: https://github.com/TBD54566975/tbdex-rs/pull/31#discussion_r1409934783

amika-sq (approved):

KendallWeihe:
cc @leordev I don't think this impacts your work, but making you aware of the use of `clippy` here

crates/crypto/src/key/public_key.rs
```diff
@@ -39,7 +38,7 @@ mod tests {
         let signature = private_key.sign(payload).unwrap();
 
         let public_key = private_key.to_public();
-        let verification_warnings = public_key.verify(&payload, &signature).unwrap();
+        let verification_warnings = public_key.verify(payload, &signature).unwrap();
```
KendallWeihe, line 41: What provoked this change? Why is `signature` passed by reference but `payload` isn't?

KendallWeihe (approved):

crates/crypto/src/key/public_key.rs
andresuribe87: This is the error https://rust-lang.github.io/rust-clippy/master/index.html#/needless_borrow

andresuribe87: BTW this was the result of running `cargo clippy --fix`

KendallWeihe: Oh I see, because `payload` was already initialized by reference whereas `signature` was not

PR #37, opened 2023-12-05
Clarify documentation of `KeyType`

andresuribe87:
From https://github.com/TBD54566975/web5-rs/pull/34 and a conversation with @decentralgabe , added documentation to clarify what I think may be confusing.

Feedback is more than welcome.

### DIFF
--- a/crates/crypto/src/key/key.rs
+++ b/crates/crypto/src/key/key.rs
@@ -2,6 +2,13 @@ use ssi_jwk::JWK;
 use ssi_jws::Error as JWSError;
 
 /// Enum defining all supported cryptographic key types.
+///
+/// Note that KeyType is NOT to be interpreted as the "kty" header defined in https://www.rfc-editor.org/rfc/rfc7517#section-4.1.
+/// Instead, "kty" (the JOSE header) should be thought of as the type for a Json Web Key (JWK)
+/// object. This enum should be thought of as the type of a cryptographic key.
+/// KeyType maps to a JWK type, but not the other way around. For example, the KeyType of "Ed25519"
+/// maps to the JWK type (i.e. the "kty" header) of "EC". The converse is not true. "EC" maps to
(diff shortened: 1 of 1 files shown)


amika-sq:
Reading these additional comments, in my opinion, make the purpose of it even more confusing 😅 

I added a ticket in the backlog to refactor this into something that more resembles what we have in Kotlin: https://github.com/TBD54566975/web5-rs/issues/38

If others agree that this helps clarify, I'm all for adding. But to me, it's bringing in a concept (JWK `kty`) that consumers of the SDK don't really need to know about.

crates/crypto/src/key/key.rs
```diff
@@ -2,6 +2,13 @@ use ssi_jwk::JWK;
 use ssi_jws::Error as JWSError;
 
 /// Enum defining all supported cryptographic key types.
+///
+/// Note that KeyType is NOT to be interpreted as the "kty" header defined in https://www.rfc-editor.org/rfc/rfc7517#section-4.1.
```
decentralgabe, line 6: this note is not necessary since this enum is not defined in the context of JOSE

crates/crypto/src/key/key.rs
```diff
@@ -2,6 +2,13 @@ use ssi_jwk::JWK;
...
+/// Instead, "kty" (the JOSE header) should be thought of as the type for a Json Web Key (JWK)
+/// object. This enum should be thought of as the type of a cryptographic key.
+/// KeyType maps to a JWK type, but not the other way around. For example, the KeyType of "Ed25519"
+/// maps to the JWK type (i.e. the "kty" header) of "EC". The converse is not true. "EC" maps to
```
decentralgabe, line 10: it's actually OKP 😉

andresuribe87:
Closing this since it's covered in https://github.com/TBD54566975/web5-rs/issues/38

PR #40, opened 2023-12-05
clippy/test fixes for main branch

amika-sq:
There was a bit of a merge race between the PR that added clippy & the Did methods PRs. Looks like a few lint issues & test compilation issues got through. 

This PR fixes those by running:
`cargo clippy --fix` 

The two linting issues were due to:
* `public_key.jwk()` is already a reference, so it doesn't need to be borrowed again
* unnecessary `return` 

The build error in tests was due to `LocalKeyManager`'s use path changing.


amika-sq:
Doh, looks like there's more merge issues than just clippy. Will add to this momentarily!

KendallWeihe (approved):

andresuribe87 (approved):

PR #44, opened 2023-12-05
Create a uniffi-bindgen crate

amika-sq:
We will be using [`uniffi-rs`](https://github.com/mozilla/uniffi-rs) to generate bindings, allowing foreign languages to be able to call into our Rust code. Our first target language will be Swift, so that tbDEX can be used on Apple platforms.

For more information on `uniffi`, check out their excellent user guide: https://mozilla.github.io/uniffi-rs/Overview.html

This PR does the setup for multi-crate workspace, as outlined in their user guide here: https://mozilla.github.io/uniffi-rs/tutorial/foreign_language_bindings.html#multi-crate-workspaces

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,3 +18,6 @@ thiserror = "1.0.50"
 serde = { version = "1.0.193", features = ["derive"] }
 serde_json = "1.0.108"
 serde_with = "3.4.0"
+
+# Update to released version once foreign traits are included (likely 0.26.0).
+uniffi = { git = "https://github.com/mozilla/uniffi-rs", rev = "7cd3aac735e905e1725d350a7a82d57aa50caaa1" }
--- a/bindings/uniffi-bindgen/Cargo.toml
+++ b/bindings/uniffi-bindgen/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "uniffi-bindgen"
+version = "0.1.0"
+edition = "2021"
+
+[[bin]]
+name = "uniffi-bindgen"
+path = "src/uniffi-bindgen.rs"
+
+[dependencies]
+uniffi = { workspace = true, features = ["cli"] }
--- a/bindings/uniffi-bindgen/src/uniffi-bindgen.rs
+++ b/bindings/uniffi-bindgen/src/uniffi-bindgen.rs
@@ -1,0 +1,3 @@
+fn main() {
+    uniffi::uniffi_bindgen_main()
+}


Cargo.toml
```diff
@@ -17,4 +17,7 @@ license-file = "LICENSE"
 thiserror = "1.0.50"
 serde = { version = "1.0.193", features = ["derive"] }
 serde_json = "1.0.108"
-serde_with = "3.4.0"
\ No newline at end of file
+serde_with = "3.4.0"
+
+# Update to released version once foreign traits are included (likely 0.26.0).
+uniffi = { git = "https://github.com/mozilla/uniffi-rs", rev = "7cd3aac735e905e1725d350a7a82d57aa50caaa1" }
```
amika-sq, line 23: Just a general note here:

Foreign trait implementation is what will allow foreign languages (e.g. Swift) to implement a Rust trait. 
For example, we have a [`KeyStore`](https://github.com/TBD54566975/web5-rs/blob/main/crates/crypto/src/key_manager/key_store/mod.rs#L11-L17) Trait, which defines the behavior that we expect of a concrete `KeyStore` implementation. Foreign trait implementation exposes bindings to allow Swift to implement that Trait, and provide it back to Rust. This will allow us to write native Swift code to store keys to the keychain!

This was initially in the current 0.25.x release, but was rolled back due to not all of the work landing before the build was cut. So they rolled it back and are going to include it in the next release. I have been using this revision for 4 weeks now, with no issues.

KendallWeihe: So does that imply that the Swift integration uses code gen? Like, Rust defines the trait, then Swift uses the bindings to read the trait definition, then codegens a foreign trait implementation by using our actual concrete implementations as the reference.

And maybe that codegen is lower-level codegen, like assembly?

amika-sq: It will make a lot more sense in the next PR where we actually do this, but here goes a summary:

We will be defining [UDL files](https://mozilla.github.io/uniffi-rs/udl_file_spec.html) that describe the interface that Rust exposes to them. Initially, this interface is going to be the absolute bare minimum required to meet our January deadline for a Swift SDK. This _will not_ include the entirety of the SDK and all of its functionality.

Uniffi reads this UDL file, and does two things:
1. Validates that the defined UDL interface is actually fulfilled. This is a compilation step.
2. Generates a C interface (known as scaffolding) into the Rust code that matches the UDL interface.

Then, once that is done, we can tell Uniffi to [generate foreign language bindings](https://mozilla.github.io/uniffi-rs/tutorial/foreign_language_bindings.html#running-uniffi-bindgen-using-a-library-file-recommended) for a target language. This is the step that actually codegens Swift code. It will only generate code that corresponds to the UDL file.

Does this generate all the code that is required to run on Swift? No. We still need to write a `KeyStore` for example, which stores private keys into the keychain. However, we DON'T need to write code to define what the `KeyStore` protocol is in Swift, as that's autogenerated by Uniffi, as it'll be defined in the UDL file. So all Swift has to do is implement the protocol, and pass it back over the language boundary back to Rust.

andresuribe87 (approved):

Cargo.toml
KendallWeihe: Interesting...

I'm confused what the Rust source code is actually for (aside from Rust apps, of course). We write the UDL file, the UDL file is a set of interface definitions, which correspond to the Rust implementations, that UDL file is used to codegen Swift interfaces, whereafter we still have to implement the Swift interface. 

Is the Rust source code only being used to validate that the UDL interface(s) correspond to an actual implementation?

As a side question too, do you know if this is the same process for all other target languages? Or is it a case-by-case basis?

amika-sq: Again, this will make a lot more sense once you see what we are actually using this for (this has been available in [this branch](https://github.com/TBD54566975/web5-rs/tree/did-jwk-restructure) for a while), or go through the uniffi user guide. We will discuss this on Discord as well.

The Rust code contains all logic. It will be shared across all foreign languages. Foreign languages, such as Swift, will have the opportunity to provide their own target-language specific functionality, if they so choose, for very specific parts of the system. 

One of those areas is `KeyStore`. We will allow target-languages to implement their own `KeyStore` instances via foreign traits. We will also allow foreign languages to just use one that's implemented in Rust (e.g. `LocalKeyManager` with an `InMemoryKeyStore`). 

Rust provides EVERYTHING ELSE about keys:
* Generating them
* Looking them up in a KeyStore
* Signing with them
* Verifying data with them

Target languages will have this provided to them, for free, implemented in Rust. They do not need to implement any of that logic, as it is exported to them via these bindings.

andresuribe87: The way I understand this: it allows dependency injection from a foreign language. Super important for secure enclaves, and trusted execution environments.

PR #65, opened 2024-04-02
Add DID Document

KendallWeihe:


crates/dids-v2/src/document.rs
```diff
@@ -0,0 +1,339 @@
+use jwk::jwk::JWK;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub enum Purpose {
```
KendallWeihe, line 5: `Purpose` is the name we used over in web5-go but I think `VerificationRelationship` is better... longer but more precise [to the spec](https://www.w3.org/TR/did-core/#verification-relationships)

PR #72, opened 2024-04-07
Replace ssi-jw* with josekit

KendallWeihe:
Driving requirement is that `josekit` supports custom signers, which will be necessary for KMS integrations

But also, `jostkit` is more widely adopted, mature, and actively maintained


KendallWeihe:
Now realizing that our `dids` crate uses the spruce dependencies which transitively depend on the `ssi-jw*` crates 🤔

KendallWeihe:
Moving to a draft, I don't like how we have the josekit JWK integrated everywhere, it's making life difficult, might introduce a `jose_jwk.rs` or something of the like and interface with all JWK related calls there

PR #74, opened 2024-04-11
Introduce `jose` crate which is a thin wrapper around the josekit open source crate

KendallWeihe:
Introduce `jose` crate such that we don't leak abstractions, and in the event we need to swap underlying implementations, then it's contained to a single crate and not everywhere

We're using `josekit` as the underlying JOSE library https://crates.io/crates/josekit

I still need to update the `credentials` crate, but will do so in a subsequent PR because this one is too unweildy already


crates/keys/src/key_manager/key_store/mod.rs
```diff
@@ -20,3 +20,6 @@ pub trait KeyStore: Send + Sync {
         private_key: Arc<dyn PrivateKey>,
     ) -> Result<(), KeyStoreError>;
 }
+
+// todo I don't think we should enable private keys to ever leave the KeyStore, 
+//      so `get` doesn't make sense and `insert` should be replaced with a `generate()` method
```
KendallWeihe, line 25: important IMO but will be for later

PR #97, opened 2024-04-19
[Draft] kendall/main into main

KendallWeihe:


nitro-neal:
I'll review-ish!

KendallWeihe:
@nitro-neal @kirahsapong be forewarned, I'm merging [this PR](https://github.com/TBD54566975/web5-rs/pull/124) in but all it consist of is taking `bindings/kt`, `bindings/swift`, and `bindings/ts` and moving them into a `binded/*` directory. This should add clarity between the distinction of "bindings" (which are Rust projects which build the binded code) and "binded" (which are non-Rust projects which contain the binded code).

```shell
rs on  kendall/binded-directory-structure [$] via 🦀 1.77.2
➜ ls binded
kt	swift	ts

rs on  kendall/binded-directory-structure [$] via 🦀 1.77.2
➜ ls bindings
uniffi	wasm
```

Also, @nitro-neal @kirahsapong I have added you two to the CODEOWNERS. I'll try to stop merging changes in while this PR is under review.

Justfile
```diff
@@ -0,0 +1,69 @@
...
+  just test-kotlin
+  just test-swift
+
+test-ts:
+  cd binded/ts && npm i && npm test
+
+test-kotlin:
+  # TODO the `sudo cp` is not what we want
+  cd binded/kt && \
+    sudo cp src/main/resources/natives/libweb5.dylib ~/Library/Java/JavaVirtualMachines/jdk-17.0.8_7.jdk/Contents/Home/bin && \
```
nitro-neal, line 63: I know you know but Just fyi if other people (like me) dont have this version we have to change this manually to run test, 

I'll keep hacking away to fix this first thing!

binded/ts/pkg/web5_wasm_bg.wasm
nitro-neal, line 1: By the way our wasm file is already 1.44 mb (and we just getting started) :) 

I have a gut feeling we will ditch js bindings and keep our oldschool web5-js in this regard

bindings/uniffi/build.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
```
nitro-neal, line 1: Can we add a comment on what this does again?

binded/ts/pkg/web5_wasm_bg.wasm
KendallWeihe: I noticed that too, but not giving up yet!

crates/jws/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+
+    pub fn from_bearer_did(
+        bearer_did: &BearerDid,
+        key_selector: &KeySelector,
+        typ: &str,
+    ) -> Result<Self, JwsError> {
+        let verification_method = bearer_did.document.get_verification_method(key_selector)?;
+        let kid = verification_method.id;
+        let alg = match verification_method.public_key_jwk.crv.as_str() {
+            "secp256k1" => "ES256K".to_string(),
```
nitro-neal, line 69: n00b rust question, but is the to_string needed?

crates/jwt/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+            Method,
+        },
+    };
+    use jws::splice_parts;
+    use keys::key_manager::local_key_manager::LocalKeyManager;
+
+    use super::*;
+    use std::sync::Arc;
+
+    #[test]
```
nitro-neal, line 106: is the end state to have tests in a different _test.rs file?

examples/KotlinExample/dev.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
nitro-neal, line 1: could be another just file? or it it not recommended to have multiple just files

examples/KotlinExample/run.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -e
+
+# TODO the `sudo cp` command below is not what we want
+(cd ../../binded/kt; \
+  sudo cp src/main/resources/natives/libweb5.dylib ~/Library/Java/JavaVirtualMachines/jdk-17.0.8_7.jdk/Contents/Home/bin; \
```
nitro-neal, line 7: Yup need to fix in this place too

nitro-neal (approved):
Awesome job! Lots of good work

crates/dids/src/method/spruce_mappers/document.rs
```diff
@@ -0,0 +1,277 @@
...
+impl VerificationMethod {
+    pub fn from_spruce(
+        spruce_verification_method: SpruceVerificationMethod,
+    ) -> Result<Self, String> {
+        match spruce_verification_method {
+            SpruceVerificationMethod::Map(ssi_vmm) => {
+                let spruce_jwk = ssi_vmm.get_jwk()?;
+                let alg = match spruce_jwk.algorithm {
+                    Some(Algorithm::ES256K) => "ES256K",
+                    Some(Algorithm::EdDSA) => "EdDSA",
```
decentralgabe, line 108: should switch out EdDSA to Ed25519 here and elsewhere
https://www.ietf.org/archive/id/draft-ietf-jose-fully-specified-algorithms-02.html

crates/jws/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+            "Ed25519" => "EdDSA".to_string(),
```
decentralgabe, line 70:
```suggestion
            "Ed25519" => "Ed25519".to_string(),
```

crates/jwt/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+
+impl Claims {
+    pub fn new(
+        issuer: Option<String>,
+        subject: Option<String>,
+        audience: Option<String>,
+        expiration: Option<i64>,
+        not_before: Option<i64>,
+        issued_at: Option<i64>,
+        jti: Option<String>,
```
decentralgabe, line 42: prob fine to exclude for now but we will need x5c

crates/jws/src/lib.rs
KendallWeihe: `"this is a str which is an array of UTF-8 bytes"` and `"this is a String type which is managed on the heap".to_string()`

basically it's a distinction between `str` and `String` types -- the former is a static memory (not sure if on the stack or in the static location in memory, but either way) and the latter is on the heap

there's a lot to even this question, just rust thangs

here's a good SO answer on this https://stackoverflow.com/a/24159933

examples/KotlinExample/dev.sh
KendallWeihe: yeah nice, done

crates/jwt/src/lib.rs
KendallWeihe: yup! will come to when needed

crates/dids/src/method/spruce_mappers/document.rs
KendallWeihe: Great callout. I did this but it'll be better to dedicate the time to do this right, so created a ticket and will follow up with it https://github.com/TBD54566975/web5-rs/issues/128

binded/ts/pkg/web5_wasm_bg.wasm
shamilovtim: Yeah like @nitro-neal said I would totally ditch it. The reality is that there is so much bespoke code in web5-js to support even the big and diverse JS runtimes such as server, browser, react native, electron that it's just not realistic at any future juncture to have Rust generate it all as WASM. In the future if we only have to support JS and Rust we'll be so far ahead of where we are today that at this point trying to hit the unrealistic goal of using WASM just doesn't seem like a great use of time.

KendallWeihe: Well said 👍

PR #126, opened 2024-04-23
Fix Kotlin libweb5.dylib loading

KendallWeihe:


KendallWeihe:
it's a bit of a hack but basically we have to edit the uniffi-generated `web5.kt` code, so I wrote a script to do it for us to keep it consistent

I think we can come to a better solution in due time but this works for now

nitro-neal (approved):

nitro-neal:
I tested and it works, nice one

Issue #127, opened 2024-04-24
Spike a larger coverage of kt bindings & integrate into existing web5-kt


KendallWeihe:
The TL;DR is: it's relatively low effort (keyword "relative" because it's still a significant effort) to build Rust bindings into every language is we don't have strong opinions on the syntactical sugar on the other side, and it's high effort to build Rust bindings and also concern ourselves with the syntactical sugar on the other side.

---

A consequence of binding Rust-to-some-other-language is that the DX expressiveness is highly constrained at the boundary. You can think of it as, every language has it's own unique set of expressive DX features, but the overlap of DX across all languages (in practical terms that means what's made available by the UniFFI UDL file) is minimal. So we can have an expressive DX in our core Rust code, which is then constricted down to a verbose DX across the binding boundary, and then we have to re-flate that expressive experience in the non-Rust language (kt in this case).

We need to make a strategic call as to whether we're going one of two routes:

1. We take what we build here and inject it into the existing https://github.com/TBD54566975/web5-kt/ project
2. We leave the old https://github.com/TBD54566975/web5-kt/ project behind and build "web5-kt" independently

(the same holds true to other languages)

The obvious choice is (1) but there is likely significantly more work to be done to achieve that relative to (2); which, we can account for, but we need to have better analysis as to the difference in effort requirements. 

For this ticket, I can spike in the ground a further build out of the Kotlin bindings, and then spend a few hours trying to use it in the existing https://github.com/TBD54566975/web5-kt/ (AKA rip out the old internals and use the binded code instead)

KendallWeihe:
I think it's safe to assert that WASM is a secondary priority and shouldn't be considered in the strategy & planning. The focus here is with respect to UniFFI, first and foremost Kotlin & Swift, but subsequently Go and Dart.

nitro-neal:
Agree with 1.

We can "piece in" the rust bindings to the existing web5-kt. put in what we have and use existing raw kt for the other

diehuxx:
Even if we decide against creating full-fledged web5-kt2 and web5-swift2, investing in language bindings will be useful in implementing net-new language implementations. We currently have no python, C#, or ruby implementations, and it's reasonable that we may want web5 or tbdex implementations for those languages in the not-too-distant future.

My worry is that we've gotten really good at implementing web5 and tbdex in different languages. Once we have a robust rust SDK and bindings infrastructure, adding new languages will be much lower lift than implementing them from scratch.

[...]

PR #130, opened 2024-04-24
Add evidence vc field

nitro-neal:
This adds the evidence field to the Verifiable Credential data model.

Since there is no concept of "any" in udl I had to make the compromise to create an evidence_json field.

It then parses it into the rust version of "any" the Value 

```
just build ✅ 
just test ✅ 
just bind ✅ 
just test-binded ✅ 
```


KendallWeihe:
This is a great ticket to cut your teeth on because of the limitations at the binding layer. As a general guiding principal, we don't want to compromise our core rust code (what's inside of `crates/`) at behest of binding. However, there is room for pragmatism, because at every abstraction layer we compound effort, so I'm in support of simplicity the further towards the core of the overall abstraction onion even if it means losing expressiveness or idiomatic patterns. 

A general rule of thumb is this: any binding solution will not support generics or generalized objects (ex. the `Object` in JavaScript), binding layers demand concrete types. And this is a real bummer because so much of the expressive DX we have in each language is a product of generics. (Side note: this is why I'm concerned about overall effort requirements if we want to build idiomatic DX in each language because we have to "reflate" the code on the other side, this matter is at the forefront of my attention https://github.com/TBD54566975/web5-rs/issues/127)

The purest Rust solution would be to use generics, so that would look something like this:

```rust
pub struct VerifiableCredential<T: Serialize + DeserializeOwned> {
    // ...other things
    pub evidence: Option<T>
}
```

But as we just learned, we can't do generics with bindings. 

Another potential pattern is to use generics at the core rust layer (within `crates/`) and then make it concrete at the binding layer (the `bindings/uniffi` crate); AKA write wrapper Rust code which bridges the gap. This is ill-advised because this can lead to a lot of Rust overhead at the binding layer which further compounds effort requirements. Instead if we can come up with something which is UniFFI-compatible at the core rust (`crates/`) layer, and it doesn't cause a complex DX on the binded side (Kotlin/Swift), then that should be our choice.

Another potential solution is to lean on the `serde_json::Value` type (as you have sort of done here), but we should be skeptical that will work well with bindings.

I really don't like the notion of using string's at the UniFFI UDL layer because it conveys zero type intent over the boundary so every instance requires the developer to have intimate knowledge to parse/stringify.

I have a proposed solution in mind, but I need to workshop it a bit more. The tools in our toolbelt are: hash maps, UDL interfaces, and traits.

decentralgabe:
I remember for earlier efforts (mine with gomobile and Adam's with rust) we had discussed having a common serialization format. In the go work we stuck to shared primitive types or JSON representations of complex structs (as bytes) between the layers. It could be that using something like JSON or protobuf acts as a nice language neutral translation layer - but can also see it adding more complexity.

agree with your statement here for the first point of exploration
> instead if we can come up with something which is UniFFI-compatible at the core rust (crates/) layer, and it doesn't cause a complex DX on the binded side (Kotlin/Swift), then that should be our choice.

though I am not opposed to the idea of a "layered solution" like
0. Pure native rust (with generics)
1. Bindings (with optional translation code to make the bindings nice)
2. Consumption of bindings
3. Idiomatic API surface in target language

PR #131, opened 2024-04-24
Add 'Any' type for JSON serializable types

KendallWeihe:
Originated from this PR https://github.com/TBD54566975/web5-rs/pull/130

We have needs for a general "any" type. The intent here is to create a solution which is compatible both in a pure rust development environment but also in a UniFFI binding project.

There is a question of: "_should we place this in `bindings/uniffi` instead? because it seems to be a constraint specific to UniFFI_"

The problem is on this line in `web5.udl`

```udl
  constructor(sequence<string> context, string id, sequence<string> type, string issuer, i64 issuance_date, i64? expiration_date, CredentialSubject credential_subject);
```

From @decentralgabe's comment here:

> 1. Pure native rust (with generics)
> 2. Bindings (with optional translation code to make the bindings nice)
> 3. Consumption of bindings
> 4. Idiomatic API surface in target language

I have had a stretch goal of limiting the burden on layer (2) to solely associated functions because leveraging the `constructor` feature in UDL means the binded language can do a traditional `VerifiableCredential(...params)`. But in order for us to leverage the `constructor` feature, the `crates/credentials/src/vc.rs` would have to inline the `Any` type in it's code. And if `Any` lives in `bindings/uniffi` crate then that means our core crate now has a dependency on the UniFFI bindings crate. Which is no bueno.

All of this to say, bindings really turns us into a pretzel. This PR is intended to spur discovery, we may choose to decline merging this. We need to establish patterns which are repeatable.

### DIFF
--- a/bindings/uniffi/Cargo.toml
+++ b/bindings/uniffi/Cargo.toml
@@ -7,6 +7,7 @@ repository.workspace = true
 license-file.workspace = true
 
 [dependencies]
+any = { path = "../../crates/any" }
 credentials = { path = "../../crates/credentials" }
 crypto = { path = "../../crates/crypto" }
 dids = { path = "../../crates/dids" }
--- a/bindings/uniffi/src/lib.rs
+++ b/bindings/uniffi/src/lib.rs
@@ -1,3 +1,4 @@
+use ::any::{Any, AnyError};
 use ::credentials::vc::{verify_vcjwt, CredentialError, CredentialSubject, VerifiableCredential};
 use ::crypto::Curve;
 use ::dids::{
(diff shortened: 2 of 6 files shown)


KendallWeihe:
So to be clear, the alternative could be to place this `Any` inside of the `bindings/uniffi` crate, and instead of this UDL code (added `...` for readability, not actually valid code)

```
  constructor(...);
```

We would add this

```
namespace web5 {
  VerifiableCredential new_verifiable_credential(..., evidence: Any);
```

KendallWeihe:
I have to write the full code to showcase the alternative. Doing so now. Moved this to a draft stage.

decentralgabe (commented):
nice - cool to prove this out

PR #132, opened 2024-04-24
Add any to uniffi crate, move VC constructor to global new_* function, add evidence to vc

KendallWeihe:
> ⚠️ **WARNING**: this is not feature complete with the VC's evidence, only exist to prove the point

### DIFF
--- a/bindings/uniffi/src/web5.udl
+++ b/bindings/uniffi/src/web5.udl
@@ -9,6 +9,18 @@ namespace web5 {
 
   [Throws=CredentialError, Async]
   VerifiableCredential verify_vcjwt([ByRef] string jwt);
+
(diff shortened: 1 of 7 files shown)


bindings/uniffi/src/web5.udl
```diff
@@ -209,7 +221,6 @@ enum CredentialError {
 };
 
 interface VerifiableCredential {
-  constructor(sequence<string> context, string id, sequence<string> type, string issuer, i64 issuance_date, i64? expiration_date, CredentialSubject credential_subject);
```
KendallWeihe, line 212: I think not-using `constructor` everywhere will probably become a rule we follow, and instead rely on the global functions (in this case `new_verifiable_credential`)

PR #156, opened 2024-05-04
Create did:dht

diehuxx:
Super un-tested, but opening for clarity and feedback on rust code style. I am not confident at all in my use of rust error handling idioms.

TODO:
- [ ] Test Vectors

### DIFF
--- a/crates/crypto/src/lib.rs
+++ b/crates/crypto/src/lib.rs
@@ -1,6 +1,11 @@
 pub mod ed25519;
 pub mod secp256k1;
 
+use std::{
+    fmt::{self, Display, Formatter},
+    str::FromStr,
+};
+
 use base64::DecodeError;
 use jwk::Jwk;
 
@@ -22,11 +27,34 @@ pub enum CryptoError {
     VerificationFailure(String),
     #[error("sign failure {0}")]
     SignFailure(String),
+    #[error("Unsupported curve {0}")]
+    UnsupportedCurve(String),
 }
 
 pub enum Curve {
     Secp256k1,
     Ed25519,
+}
+
+impl FromStr for Curve {
+    type Err = CryptoError;
+
+    fn from_str(input: &str) -> Result<Curve, CryptoError> {
+        match input {
+            "Ed25519" => Ok(Curve::Ed25519),
+            "secp256k1" => Ok(Curve::Secp256k1),
+            _ => Err(CryptoError::UnsupportedCurve(input.to_string())),
+        }
+    }
+}
+
+impl Display for Curve {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            Curve::Ed25519 => write!(f, "Ed25519"),
+            Curve::Secp256k1 => write!(f, "secp256k1"),
+        }
+    }
 }
 
 pub trait CurveOperations {
--- a/crates/dids/Cargo.toml
+++ b/crates/dids/Cargo.toml
@@ -7,18 +7,24 @@ repository.workspace = true
 license-file.workspace = true
 
 [dependencies]
+base64 = { workspace = true }
+byteorder = "1.5.0"
+chrono = { workspace = true }
 crypto = { path = "../crypto" }
 did-jwk = "0.1.1"
 did-web = "0.2.2"
 jwk = { path = "../jwk" }
 keys = { path = "../keys" }
+pkarr = "1.1.5"
 regex = "1.10.4"
+reqwest = { version = "0.11", features = ["blocking"] }
 serde = { workspace = true }
 serde_json = { workspace = true }
 ssi-core = "0.1.0"
 ssi-dids = "0.1.1"
 ssi-jwk = "0.1.2"
 thiserror = { workspace = true }
+zbase32 = "0.1.2"
 
 [dev-dependencies]
 chrono = { workspace = true }
--- a/crates/dids/src/method/dht/bep44.rs
+++ b/crates/dids/src/method/dht/bep44.rs
@@ -11,18 +11,25 @@ where
 {
     let microseconds_since_epoch = SystemTime::now()
         .duration_since(UNIX_EPOCH)
-        .map_err(|_| MethodError::DidCreationFailure("Could not get time since epoch".to_string()))?
+        .map_err(|_| {
+            MethodError::DidPublishingFailure("Could not get time since epoch".to_string())
+        })?
         .as_micros();
 
     let mut signable =
         format!("3:seqi{}e1:v{}:", microseconds_since_epoch, message.len()).into_bytes();
     signable.extend(message);
 
-    let signature = sign(signable)
-        .map_err(|_| MethodError::DidCreationFailure("Failed to sign BEP44 message".to_string()))?;
+    let signature = sign(signable).map_err(|_| {
+        MethodError::DidPublishingFailure("Failed to sign BEP44 message".to_string())
+    })?;
 
     let mut seq_bytes = vec![];
-    seq_bytes.write_u128::<BigEndian>(microseconds_since_epoch);
+    seq_bytes
+        .write_u128::<BigEndian>(microseconds_since_epoch)
+        .map_err(|_| {
+            MethodError::DidPublishingFailure("Failed to write big endian seq".to_string())
+        })?;
 
     let mut encoded = Vec::new();
     encoded.extend(signature);
--- a/crates/dids/src/method/dht/bep44.rs
+++ b/crates/dids/src/method/dht/bep44.rs
@@ -1,0 +1,33 @@
+use std::time::{SystemTime, UNIX_EPOCH};
+
+use byteorder::{BigEndian, WriteBytesExt};
+use keys::key_manager::KeyManagerError;
+
+use crate::method::MethodError;
+
+pub fn encode_bep44_message<F>(message: &[u8], sign: F) -> Result<Vec<u8>, MethodError>
+where
+    F: Fn(Vec<u8>) -> Result<Vec<u8>, KeyManagerError>,
+{
+    let microseconds_since_epoch = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .map_err(|_| MethodError::DidCreationFailure("Could not get time since epoch".to_string()))?
+        .as_micros();
+
+    let mut signable =
+        format!("3:seqi{}e1:v{}:", microseconds_since_epoch, message.len()).into_bytes();
+    signable.extend(message);
+
+    let signature = sign(signable)
+        .map_err(|_| MethodError::DidCreationFailure("Failed to sign BEP44 message".to_string()))?;
+
+    let mut seq_bytes = vec![];
+    seq_bytes.write_u128::<BigEndian>(microseconds_since_epoch);
+
+    let mut encoded = Vec::new();
+    encoded.extend(signature);
+    encoded.extend(seq_bytes);
+    encoded.extend(message);
+
+    Ok(encoded)
+}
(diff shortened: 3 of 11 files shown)


crates/dids/src/method/dht/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+            Some("0".to_string()), // PR review: i copied this from jwk.rs. do we want this to be None?
+        )?;
+        let public_key = key_manager.get_public_key(&key_alias)?;
+
+        // Create did uri
+        let public_jwk = public_key.jwk()?;
+        let jwk_string = serde_json::to_string(public_jwk.as_ref()).map_err(|_| {
+            MethodError::DidCreationFailure("failed to serialize public jwk".to_string())
+        })?;
+        let identifier = zbase32::encode_full_bytes(jwk_string.as_bytes());
```
KendallWeihe, line 63: as far as I can tell the spec would call this either a suffix or identity_key

```suggestion
        let suffix = zbase32::encode_full_bytes(jwk_string.as_bytes());
```

or 

```suggestion
        let identity_key = zbase32::encode_full_bytes(jwk_string.as_bytes());
```

crates/dids/src/method/dht/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+        // Generate private and public keypair
+        let key_alias = key_manager.generate_private_key(
+            Curve::Ed25519,
+            Some("0".to_string()), // PR review: i copied this from jwk.rs. do we want this to be None?
+        )?;
+        let public_key = key_manager.get_public_key(&key_alias)?;
+
+        // Create did uri
+        let public_jwk = public_key.jwk()?;
+        let jwk_string = serde_json::to_string(public_jwk.as_ref()).map_err(|_| {
```
KendallWeihe, line 60: either way works, either are fine with me

```suggestion
        let jwk_string = serde_json::to_string(&public_jwk).map_err(|_| {
```

KendallWeihe: I kinda feel like we'd be wiser to expand our `MethodError` enum set to include more specific errors, and then write an `impl From` ([like we do here](https://github.com/TBD54566975/web5-rs/blob/b847389b5ec619583b643ca59a2d54ac87f11e65/crates/jws/src/lib.rs#L33)) and then we wouldn't have to call `.map_err`

I have nothing against `map_err` so that doesn't really matter either way, but I do think it's good practice to have error enums which are semantically relevant rather than relying on string messages... I'm going to open a ticket for this. no need to address right now unless you feel choose to.

KendallWeihe: https://github.com/TBD54566975/web5-rs/issues/191

KendallWeihe: Edited the original comment, as that suggestion is incorrect. Apologies.

crates/dids/src/method/dht/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+    const NAME: &'static str = "dht";
+
+    fn create(
+        key_manager: Arc<dyn KeyManager>,
+        options: DidDhtCreateOptions,
+    ) -> Result<BearerDid, MethodError> {
+        // Generate private and public keypair
+        let key_alias = key_manager.generate_private_key(
+            Curve::Ed25519,
+            Some("0".to_string()), // PR review: i copied this from jwk.rs. do we want this to be None?
```
KendallWeihe, line 54: this is correct BTW, and this is a new paradigm only made available in rust

which is, the `KeyManager` trait's `generate_private_key` enables the caller to override setting the `key_alias` (in the case of did:dht and did:jwk, both have an ID of `0`), which is necessary for subsequently instantiating a bearer DID from an existing key manager in order for the Verification Method's key ID to match the key alias in the key manager

KendallWeihe: this is made possible in other SDKs through the portable DID, but I'm making it an initiative to second-class the serialization of private key material and instead make secure enclave the first class assumption (which is to say, keys are never serialized)

crates/dids/src/method/dht/mod.rs
KendallWeihe: fyi the [Terminology in the spec](https://did-dht.com/#terminology) was ambiguous on this matter but it's been iron'd out -- the proper variable name here is `suffix`

crates/dids/src/method/dht/mod.rs
KendallWeihe: FYI, we should do this anyways https://github.com/TBD54566975/web5-rs/pull/199

crates/crypto/src/lib.rs
```diff
@@ -22,13 +27,36 @@ pub enum CryptoError {
...
+    #[error("Unsupported curve {0}")]
+    UnsupportedCurve(String),
 }
 
 pub enum Curve {
     Secp256k1,
     Ed25519,
 }
 
+impl FromStr for Curve {
```
KendallWeihe, line 39: yes! https://github.com/TBD54566975/web5-rs/issues/205

crates/dids/Cargo.toml
```diff
@@ -7,18 +7,24 @@ repository.workspace = true
...
+byteorder = "1.5.0"
+chrono = { workspace = true }
 crypto = { path = "../crypto" }
 did-jwk = "0.1.1"
 did-web = "0.2.2"
 jwk = { path = "../jwk" }
 keys = { path = "../keys" }
+pkarr = "1.1.5"
 regex = "1.10.4"
+reqwest = { version = "0.11", features = ["blocking"] }
```
KendallWeihe, line 20: if I'm interpreting [this](https://docs.rs/reqwest/0.12.4/reqwest/#wasm) correctly, `blocking` isn't supported in wasm

> The Client implementation automatically switches to the WASM one when the target_arch is wasm32, the usage is basically the same as the async api. Some of the features are disabled in wasm : [tls](https://docs.rs/reqwest/0.12.4/reqwest/tls/index.html), [cookie](https://docs.rs/reqwest/0.12.4/reqwest/cookie/index.html), [blocking](https://docs.rs/reqwest/0.12.4/reqwest/blocking/index.html).

That's still okay with me if we want to stick with it in this PR, but @diehuxx if you agree with my interpretation of the matter, could you open a ticket on the backlog to explore implications & alternatives.

KendallWeihe: (wasm is a non-priority right now but I want to earmark it because I suspect wasm will be a priority at some point)

crates/dids/src/method/dht/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+
+        // Create did uri
+        let public_jwk = public_key.jwk()?;
+        let jwk_string = serde_json::to_string(public_jwk.as_ref()).map_err(|_| {
+            MethodError::DidCreationFailure("failed to serialize public jwk".to_string())
+        })?;
+        let suffix = zbase32::encode_full_bytes(jwk_string.as_bytes());
+        let uri = format!("did:dht:{}", suffix);
+        let identifier = Identifier::parse(&uri).map_err(|e| {
+            MethodError::DidCreationFailure(format!("Failed to parse did:jwk uri {} {}", &uri, e))
```
KendallWeihe, line 60:
```suggestion
            MethodError::DidCreationFailure(format!("Failed to parse did:dht uri {} {}", &uri, e))
```

on a similar note, you could automatically map the error by adding this to the `IdentifierError`

```rust
    #[error(transparent)]
    IdentifierError(#[from] IdentifierError),
```

also if you're curious, go check out `jws/src/lib.rs` and you'll see this for mapping serde serialization errors (a few lines above this line)

```rust
impl From<SerdeJsonError> for JwsError {
    fn from(err: SerdeJsonError) -> Self {
        JwsError::SerdeJsonError(err.to_string())
    }
}
```

crates/dids/src/method/dht/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+        let public_key = key_manager.get_public_key(&key_alias)?;
+
+        // Create did uri
+        let public_jwk = public_key.jwk()?;
+        let jwk_string = serde_json::to_string(public_jwk.as_ref()).map_err(|_| {
+            MethodError::DidCreationFailure("failed to serialize public jwk".to_string())
+        })?;
+        let suffix = zbase32::encode_full_bytes(jwk_string.as_bytes());
+        let uri = format!("did:dht:{}", suffix);
+        let identifier = Identifier::parse(&uri).map_err(|e| {
```
KendallWeihe, line 59: you may want to inline this, like so 

```suggestion
        let identifier = Identifier::parse(&format!("did:dht:{}", suffix)).map_err(|e| {
```

...and subsequently use the `identifier.uri` as the source of truth, that way we're fixing the `Identifier` if we discover issues, which will propagate to all uses of it

crates/dids/src/method/dht/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+                })?;
+
+            let bep44_message = encode_bep44_message(&packet_bytes, |payload| {
+                key_manager.sign(&key_alias, &payload)
+            })?;
+
+            // HTTP PUT
+            let client = Client::new();
+            let response = client
+                .put(DEFAULT_GATEWAY_URL)
```
KendallWeihe, line 101: looks like [`/:id` param is required to be](https://did-dht.com/#put) in the URL path (but that may be a new development which is not currently supported in our test vectors and/or deployed gateways?)

so basically:

```suggestion
                .put(format!("{}/{}", DEFAULT_GATEWAY_URL, suffix))
```

crates/dids/src/method/dht/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+
+            let packet_bytes = packet
+                .build_bytes_vec_compressed()
+                .map_err(|_| -> MethodError {
+                    MethodError::DidPublishingFailure(
+                        "Failed to serialize DNS packet to bytes".to_string(),
+                    )
+                })?;
+
+            let bep44_message = encode_bep44_message(&packet_bytes, |payload| {
```
KendallWeihe, line 94: nice! a `sign` callback parameter... I just created https://github.com/TBD54566975/web5-rs/issues/208 which explores using this pattern everywhere instead of the existing `bearer_did`+`key_selector` design.

crates/dids/src/method/dht/bep44.rs
```diff
@@ -0,0 +1,40 @@
...
+use byteorder::{BigEndian, WriteBytesExt};
+use keys::key_manager::KeyManagerError;
+
+use crate::method::MethodError;
+
+pub fn encode_bep44_message<F>(message: &[u8], sign: F) -> Result<Vec<u8>, MethodError>
+where
+    F: Fn(Vec<u8>) -> Result<Vec<u8>, KeyManagerError>,
+{
+    let microseconds_since_epoch = SystemTime::now()
```
KendallWeihe, line 12: I believe this needs to be seconds, not micro seconds

> seq MUST be set to the current [Unix Timestamp](https://did-dht.com/#term:unix-timestamp) in seconds

crates/dids/src/method/dht/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+
+        if let Some(vms) = options.verification_methods {
+            verification_methods.extend(vms);
+        }
+
+        let document = Document {
+            id: uri.clone(),
+            verification_method: verification_methods,
+            service: options.service,
+            ..Default::default()
```
KendallWeihe, line 79: [from the spec](https://did-dht.com/#identity-key)

> Identity Keys MUST have the [Verification Relationships](https://did-dht.com/#verification-relationships) Authentication, Assertion, Capability Invocation, and Capability Delegation.

we need to go ahead and create those four verification relationships in the document creation

PR #157, opened 2024-05-05
feat: separate DID Method trait from Create trait for create-able DIDs

enmand:
Some DID methods allow for DID creation (e.g. with some key pair) and some have more infrastructure requirements around them (like `did:web`). This PR separates the trait for "create-able" DIDs, so that DID methods like `did:web` don't have to carry around trait types and no-implementation-implementations for trait functions they don't need.

### DIFF
--- a/crates/credentials/src/vc.rs
+++ b/crates/credentials/src/vc.rs
@@ -133,7 +133,7 @@ mod test {
         document::VerificationMethodType,
         method::{
             jwk::{DidJwk, DidJwkCreateOptions},
-            Method,
+            Create,
         },
     };
     use keys::key_manager::local_key_manager::LocalKeyManager;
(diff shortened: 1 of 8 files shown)


enmand:
On second though, maybe there is value in having all the methods defined in a single trait, even if it's a bit cumbersome for some method implementations -- I could see a world where there are separate e.g. `Resolve`, `Create`, `Update` and `Deactivate` trait but I also see value in having all the methods under a single trait

KendallWeihe:
I'm a little torn on this and need to think about it. I think it's a great idea to come to an implementation wherein the lack of support of a feature is determinable at build-time, as opposed to how we currently have things which wouldn't be discovered until runtime. However, strictly aligning to [the specification](https://www.w3.org/TR/did-core/#method-operations) I view all of the various CRUD operations on a method as a bundle of features and so at face value it makes sense for them to be in a single trait. I think... probably what you have proposed here @enmand is ideal, but let's let it sit for a bit and we can contemplate 🤔

KendallWeihe (approved):
let's do it!

Issue #153, opened 2024-05-03
did:jwk creation missing verification relationships


KendallWeihe:
From here https://github.com/TBD54566975/web5-rs/blob/f74a68429159881c5c77d55886b8c39fbb9fcfe1/crates/dids/src/method/jwk.rs#L60

When `DidJwk::create()` is called, the appropriate verification relationships are missing. For example, refer to [the test vector](https://github.com/TBD54566975/web5-spec/blob/cb6621177190b08ff38be7a1a4ab84e4a6cbcca7/test-vectors/did_jwk/resolve.json#L30) and notice `authentication`, `assertionMethod`, `capabilityInvocation` and `capabilityDelegation` are all set with one value equal to the `verificationMethod[0].id`.

KendallWeihe:
this was completed in #163

Issue #146, opened 2024-05-03
Update GitHub CI workflow to use Hermit dependencies


KendallWeihe:
Recently did this over in `tbdex-rs` https://github.com/TBD54566975/tbdex-rs/pull/45

We should use our hermit dependencies in the CI pipeline as to maintain consistency across environments

KendallWeihe:
Let's also make sure we include workflow job(s) for executing the bindings; we don't want PRs merged which break bindings

KendallWeihe:
Make sure CI workflows run on all PRs even if they aren't only into `main`; I realized that's not how it currently is when I opened #186

PR #163, opened 2024-05-06
update jwk method for creation missing verification relationships

nitro-neal:
When DidJwk::create() is called, the appropriate verification relationships are missing. 

authentication, assertionMethod, capabilityInvocation and capabilityDelegation are missing and need to be set with one value equal to the verificationMethod[0].id.

closes - https://github.com/TBD54566975/web5-rs/issues/153


KendallWeihe (approved):
nice!

KendallWeihe:
@nitro-neal you can run `just lint` and it'll fix the formatting issue (currently causing CI to fail; the pipeline stuff will soon change with #146)

Issue #171, opened 2024-05-07
Consider combining `jws` and `jwt` crates into one crate, `jose`


KendallWeihe:
[This comment](https://github.com/TBD54566975/web5-rs/issues/144#issuecomment-2098505777) got me thinking about how JWE's fit into our approach. Currently we only have a use case for JWS but it's safe to assume we'll need to include JWE in the future.

## Guiding Principal

We’re not building generic JOSE solutions, we’re building JOSE solutions which are specific to our use cases, and specifically how DID's are a novel public/private key cryptography solution, however we also want to design the surface such that it’s extensible if our use cases broaden.

## Proposal

Create a new crate called `jose` which will have the following modules:

- `jwt/`
    - `jwe.rs` // empty, stubbed in to prove the case for extensibility
    - `jws.rs`
- `headers.rs` // jws & jwe both use "JOSE Headers"
- `jwe.rs` // empty, stubbed in to prove the case for extensibility
- `jws.rs`

The major caveat here is we maintain `jwk` as an independent crate. "_But Kendall, JWK is a part of the JOSE space!_" Yes, but, by including JWK's in our `jose` crate creates a circular dependency problem. Here's a visual to showcase the dependency problem.

<img width="327" alt="Screenshot 2024-05-07 at 10 36 31 AM" src="https://github.com/TBD54566975/web5-rs/assets/3602993/3a014985-2aac-48a1-aa54-8bf9f6608c19">

(#142 covers the discussion of broadening our key representation to more than solely JWK's)

There are so many different ways to do this, and all with flaws, and we could bike shed forever

KendallWeihe:
The more I think about this, the less I like it. I think at the rust core layer we want to maximize modularity. I'll let this ticket sit for a bit to allow others to chime in, but expect to close it as won't implement.

KendallWeihe:
Closing as won't-do

Issue #150, opened 2024-05-03
Use #144 in `credentials/vc.rs`


KendallWeihe:
Given we rework `jwt` in accordance with #144 and `jws` in accordance with #143 then we'll need to update `vc.rs` to use the latest.

A few items come to mind, but the work may span beyond this:
- `JwsHeader::from_bearer_did` will no longer exist (given we remove in in accordance with the commentary [linked here](https://github.com/TBD54566975/web5-rs/issues/143#issuecomment-2093593691))
- we can use `None` for the `header` param in `Jwt::sign()` as the default is acceptable
- no need to encode the claims, as that will happen internal to `Jwt::sign()`

KendallWeihe:
Picking up this work and some things are coming to mind:

- we should probably codify the differentiation between VC Data Model 1.1 and 2.0
- are there differences in the ways we should implement the JWT across data models?

KendallWeihe:
@decentralgabe could you answer or direct me to resources to answer this question?

> are there differences in the ways we should implement the JWT across data models?

KendallWeihe:
I created a new ticket for the two above comments https://github.com/TBD54566975/web5-rs/issues/175

@decentralgabe please use the new ticket for discussions on the matter

decentralgabe:
I will respond in 175 as well, but yes there is a difference between 1.1 and 2.0

1.1 follows unfortunate mapping rules defined here https://www.w3.org/TR/vc-data-model/#jwt-encoding

2.0 follows standards jwt conventions

Issue #144, opened 2024-05-03
Refactor `jwt` to be consistent with #143


KendallWeihe:
#143 proposes a new design for the `jws` crate; basically, we'll want to do the same with the `jwt` crate as the two are semantically mostly similar

So the new `jwt` design would be something like this:

```rust
use dids::{bearer::BearerDid, document::KeySelector};
use jws::v2::Header;
use serde::Serialize;

pub enum Error {
    // error cases
}

pub trait Claims: Serialize {}

pub struct BaseClaims {
    // all of the existing claims
}

pub struct Decoded<T: Claims> {
    pub header: Header,
    pub claims: T,
    pub signature: String,
    pub parts: Vec<String>,
}

pub struct Jwt;

impl Jwt {
    pub fn sign<T: Claims>(
        bearer_did: &BearerDid,
        key_selector: &KeySelector,
        header: Option<Header>,
        claims: &T,
    ) -> Result<String, Error> {
        unimplemented!()
    }

    pub fn decode<T: Claims>(compact_jws: &str) -> Result<Decoded<T>, Error> {
        unimplemented!()
    }

    pub fn verify<T: Claims>(compact_jws: &str) -> Result<Decoded<T>, JwtError> {
        unimplemented!()
    }
}
```

As with #143, this will break `credentials` and perhaps other things, so we should be conscious about the way the work is split up and reviewable

KendallWeihe:
I'm actually not 100% about this bit, need to consider the VC dependency on JWT as a way to consider the ideal pattern.

```rust
pub trait Claims {
    fn encode(&self) -> String;
}
```

The idea being, we want a way for consuming projects (i.e. `credentials` crate) to be able to extend the `BaseClaims` to add. Also, we don't need to over engineer this; if the possible set of claims for all of our use cases is well known then we can just define all of the possible claims in the `jwt` crate itself, no need to get fancy if it's not required.

KendallWeihe:
[...]

Issue #143, opened 2024-05-03
Refactor `jws` to be aligned with specification concepts


KendallWeihe:
I take issue with a number of things inside the `jws` crate

- Duplicate signing function
- `new_from_encoded()` and `encode()` both only used internally (albeit `encode()` is used in `vc.rs` but that needs to be changed (TODO working on another ticket for refactoring the `jwt` crate) -- both of these could be removed
- we're signing a "JwsHeader" but that doesn't really make sense because what's being signed is both the header and the payload
- `splice_parts` is out on it's own, and technically is splicing a **compact** JS
- we have `JwsError` but it's the only error in the crate, so why not just call it `Error`?

I would propose we rethink this crate altogether with the following design:

(I'm going to also include the "decode" design from this ticket https://github.com/TBD54566975/web5-rs/issues/118)

```rust
use dids::{bearer::BearerDid, document::KeySelector};

pub enum Error {
    // ... error cases
}

pub struct Header {
    pub alg: String,
    // ... more properties
}

impl Header {
    pub fn from_bearer_did(
        bearer_did: &BearerDid,
        key_selector: &KeySelector,
        typ: &str,
    ) -> Result<Self, Error> {
        unimplemented!()
    }
}

pub struct Decoded {
    pub header: Header,
    pub payload: Vec<u8>,
    pub signature: String,
    pub parts: Vec<String>,
}

pub struct CompactJws;

impl CompactJws {
    pub fn sign(
        bearer_did: &BearerDid,
        key_selector: &KeySelector,
        header: &Header,
        encoded_payload: &[u8], // NOTE: this is not base64 encoded, that happens internal to this function
    ) -> Result<String, Error> {
        unimplemented!()
    }

    pub fn decode(compact_jws: &str) -> Result<Decoded, Error> {
        unimplemented!()
    }

    pub fn verify(compact_jws: &str) -> Result<Decoded, Error> {
        unimplemented!()
    }
}
```

These changes would break at least `jwt` and `credentials` so if we choose to move forward with this proposal then we should probably string these work items together

nitro-neal:
why not keep it `JwsError` and we can have at least one default error for every class?

and yes! I like this proposed way very much, very similar to web5-go right?

not sure why we need this function:

```
impl Header {
    pub fn from_bearer_did(
        bearer_did: &BearerDid,
        key_selector: &KeySelector,
        typ: &str,
    ) -> Result<Self, Error> {
        unimplemented!()
    }
}
```

[...]

PR #173, opened 2024-05-07
Refactor jwt #144

KendallWeihe:
# Background

This is a part of a multi-PR (& multi-crate) initiative to complete #143 #144 & #150 without breaking anything, whereafter we'll come through and remove old code. This is the reasoning for `v2` in the file name `lib_v2.rs` -- as to not break anything but to see through the initiative in reviewable PRs. This PR addresses #144. 

# Proposed changes

Notice [the comment here](https://github.com/TBD54566975/web5-rs/issues/144#issuecomment-2098505777) -- after speaking w/ @decentralgabe it's clear we have a requirement to structure the code such that we can add support for JWE at a future date, ideally without inducing breaking changes. I've moved forward with option (3) from that comment, to add future extensibility via namespacing the JWT implementation for JWS's. This is why the `Jwt` implementation is inside of the `jws.rs` file, and why there is an empty `jwe.rs` file; at a future date we can add JWT's as JWE inside of the `jwe.rs` file.

The implementation of JWT's as JWS piggy backs off of the new feature set in the `jws` crate made available from #162.

# Additional work items

Following the work here, we'll implement #150 and subsequently we'll go through and remove all of the old ("non-v2") code.

I've also opened a new ticket to discuss a design change around our rust crate approach #171


crates/jwt/src/lib_v2.rs
```diff
@@ -0,0 +1,42 @@
...
+impl From<SerdeJsonError> for JwtError {
+    fn from(err: SerdeJsonError) -> Self {
+        JwtError::SerdeJsonError(err.to_string())
+    }
+}
+
+pub trait Claims: Serialize + DeserializeOwned {}
+
+#[derive(Serialize, Deserialize, Debug, Default)]
+pub struct BaseClaims {
```
KendallWeihe, line 25: I want to rename this to `RegisteredClaims` as to match the spec

crates/jwt/src/jwe.rs
```diff
@@ -0,0 +1 @@
+// implementation for JWT's as JWE's
```
KendallWeihe, line 1: torn on if we should leave this or remove it, I'm leaning towards remove it

crates/jwt/src/lib_v2.rs
KendallWeihe: done

crates/jwt/src/jwe.rs
KendallWeihe: decided to stub in scaffolded code, but not implement and instead created ticket for subsequent work #174

nitro-neal:
this is great, but if the complexity gets too high or there is too much duplicate code it may be better just to do this:

```
One struct with differently named functions Jwt::sign_jws()/Jwt::sign_jwe()/etc (better IMO)
```

In js can do something like this, I don't think rust can do this: Jwt::sign(jweClaim | jwsClaims)

Another approach is have Jwt::sign() which is default a jws (because like 90% of things will use this probabbly)

but have an optional flag or options object that can change it to jwe

Just some thoughts! But I'm fine with the current impl

KendallWeihe:
honestly, I kinda like `Jwt::sign_jws()` and `Jwt::sign_jwe()` better than how it is now with the namespace `jwt::jwe::Jwt` 

opened a ticket for it https://github.com/TBD54566975/web5-rs/issues/177

nitro-neal (approved):

Issue #175, opened 2024-05-08
Consider differentiating between VC Data Models 1.1 and 2.0


KendallWeihe:
Originally spurred during the work on #150. We currently have `crates/credentials/src/vc.rs` and that implement VC Data Model 1.1, but we should consider how VC Data Model 2.0 will fit into our code. Furthermore, are the JWT implementations for each distinct or is there overlap?

decentralgabe:
v2.0 will be more straightforward -- there is no mapping needed. the semantics of each layer are preserved. https://w3c.github.io/vc-jose-cose/#securing-json-ld-verifiable-credentials-with-jose

there is guidance that you should not conflict...like

> Implementers SHOULD avoid setting JWT claims to values that conflict with the values of verifiable credential properties when a claim and property pair refer to the same conceptual entity, especially with pairs such as iss and issuer, jti and id, and sub and credentialSubject.id. For example, JWK claim iss should not be set to a value which conflicts with the value of verifiable credential property issuer.

1.1 follows unfortunate mapping rules defined here https://www.w3.org/TR/vc-data-model/#jwt-encoding

this gives us optionality to include or delete properties that overlap between the layers (e.g. issuer/iss or id/jti) and so far we've chosen to remove them from the VC during JWT encoding and rely on the JWT values.

PR #176, opened 2024-05-08
Use latest jws & jwt functionality, add VC verify and decode functions to match #150

KendallWeihe:
# Background

This is a part of a messy multi-step multi-PR campaign to revamp JWS (#143), JWT (#144) and the uses in VC (#150); this PR addresses #150. I will subsequently open a PR removing all the non-v2 code and update bindings accordingly.

# Changes

- Use the latest in the `jwt` crate (changes in #173) -- this includes `RegisteredClaims`, and the new `sign`/`verify`/`decode` functions
- Added `VerifiableCredential::verify()` and `VerifiableCredential::decode()` functions, as to match the paradigm with the latest in `jws` and `jwt`

# Forward looking

- I'm not entirely sold on the way I have implemented the `verify` and `decode` so we can iterate on the approach there
- the same goes for `VcJwtClaims`
- Created this ticket to consider how we can support both VC Data Model 1.1 and 2.0 https://github.com/TBD54566975/web5-rs/issues/175 -- both of the above bullet points may be dependent upon how we think about supporting both data models
- in a subsequent PR I will remove all of the old code, and update bindings accordingly

### DIFF
--- a/crates/credentials/src/vc.rs
+++ b/crates/credentials/src/vc.rs
@@ -1,7 +1,9 @@
-use base64::{engine::general_purpose, Engine as _};
 use dids::{bearer::BearerDid, document::KeySelector};
-use jws::{splice_parts, JwsError, JwsHeader};
-use jwt::{sign_jwt, verify_jwt, Claims, JwtError};
+use jws::v2::JwsError;
+use jwt::{
+    jws::Jwt,
+    lib_v2::{Claims, JwtError, RegisteredClaims},
+};
 use serde::{Deserialize, Serialize};
 use std::{collections::HashMap, sync::Arc};
 
@@ -72,9 +74,8 @@ impl VerifiableCredential {
         bearer_did: &BearerDid,
         key_selector: &KeySelector,
     ) -> Result<String, CredentialError> {
-        let header = JwsHeader::from_bearer_did(bearer_did, key_selector, "JWT")?;
         let claims = VcJwtClaims {
-            base_claims: Claims {
+            registered_claims: RegisteredClaims {
                 issuer: Some(self.issuer.clone()),
                 jti: Some(self.id.clone()),
                 subject: Some(self.credential_subject.id.clone()),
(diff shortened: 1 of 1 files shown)


decentralgabe (approved):

crates/credentials/src/vc.rs
```diff
@@ -85,45 +86,40 @@ impl VerifiableCredential {
             vc: self.clone(),
         };
 
-        let encoded_header = header.encode()?;
-        let encoded_claims = claims.encode()?;
+        let jwt = Jwt::sign(bearer_did, key_selector, None, &claims)?;
```
nitro-neal, line 89: Eventually I think we will want a concept of a vcJwt as it is not exactly the same as a jwt

nitro-neal (approved):

crates/credentials/src/vc.rs
```diff
@@ -85,45 +86,40 @@ impl VerifiableCredential {
...
+    pub async fn verify(jwt: &str) -> Result<Self, CredentialError> {
+        let jwt_decoded = Jwt::verify::<VcJwtClaims>(jwt).await?;
+
+        // TODO Implement semantic VC verification rules https://github.com/TBD54566975/web5-rs/issues/151
+
+        Ok(jwt_decoded.claims.vc)
+    }
+
+    pub fn decode(jwt: &str) -> Result<Self, CredentialError> {
+        let jwt_decoded = Jwt::decode::<VcJwtClaims>(jwt)?;
```
nitro-neal, line 103: Ah perfect

crates/credentials/src/vc.rs
nitro-neal: for reference :

https://github.com/TBD54566975/web5-js/blob/main/packages/credentials/src/verifiable-credential.ts#L117C1-L134C4
```
  public async sign(options: VerifiableCredentialSignOptions): Promise<string> {
    const vcJwt: string = await Jwt.sign({
      signerDid : options.did,
      payload   : {
        vc  : this.vcDataModel,
        nbf : Math.floor(new Date(this.vcDataModel.issuanceDate).getTime() / 1000),
        jti : this.vcDataModel.id,
        iss : options.did.uri,
        sub : this.subject,
        iat : Math.floor(Date.now() / 1000),
        ...(this.vcDataModel.expirationDate && {
          exp: Math.floor(new Date(this.vcDataModel.expirationDate).getTime() / 1000),
        }),
      }
    });

    return vcJwt;
  }
  ```

PR #179, opened 2024-05-08
Specify rust version in `just setup` recipe

### LINKED ISSUE #178: Specify a specific version for the rustup setup recipe
We have this Justfile recipe:

```justfile
# Setup local development environment
setup:
  #!/bin/bash
    rustup default stable
  fi
```

But we should specify a specific version in the code, because otherwise if people run the recipe at different points in time they may be using different versions of the underlying dependency

KendallWeihe:
Closes #178 

---

Remove your `.hermit/` directory and re-run `just setup` if you want to update your environment to the latest

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -4,7 +4,7 @@ set shell := ["bash", "-uc"]
 setup:
   #!/bin/bash
   if [[ "$(cargo 2>&1)" == *"rustup could not choose a version of cargo to run"* ]]; then
-    rustup default stable
+    rustup default 1.78.0
   fi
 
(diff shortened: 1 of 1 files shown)


nitro-neal (approved):

nitro-neal:
we will need to run just setup again right?

KendallWeihe:
> we will need to run just setup again right?

yeah, I doubt we'll run into any issues across across the versions we may have but it's good practice

```shell
rm -rf .hermit && just setup
```

PR #182, opened 2024-05-08
implement vc verification

nitro-neal:
This pr does a change on how we process and verify VCs from a VcJwt:

This looks at registered claims fields and does a few checks:
* If a required register claim does not exist, throw an error
* If they don't exist in the underlying claims.vc field, then use the registers claims as the new value for the vc
* If the VC had a value, make sure it matches. If there is a missmatch, throw an error

Here is the spec specific text:
```
   * Once the verifications are successful, when recreating the VC data model object, this function will:
   * - If exp is present, the UNIX timestamp MUST be converted to an [XMLSCHEMA11-2] date-time, and MUST be used to set the value of the expirationDate property of credentialSubject of the new JSON object.
   * - If iss is present, the value MUST be used to set the issuer property of the new credential JSON object or the holder property of the new presentation JSON object.
   * - If nbf is present, the UNIX timestamp MUST be converted to an [XMLSCHEMA11-2] date-time, and MUST be used to set the value of the issuanceDate property of the new JSON object.
   * - If sub is present, the value MUST be used to set the value of the id property of credentialSubject of the new credential JSON object.
   * - If jti is present, the value MUST be used to set the value of the id property of the new JSON object.
```

   

for reference:

This "minimum" vc object is valid according to the spec
```
{
  "vc": {
    "@context": [
      "https://www.w3.org/2018/credentials/v1"
    ],
    "type": [
      "VerifiableCredential"
    ],
    "credential_subject": {
    }
  },
  "iss": "did:jwk:eyJhbGciOiJFZERTQSIsImNydiI6IkVkMjU1MTkiLCJrdHkiOiJPS1AiLCJ4IjoiRE5NdDhrYUNPanlSVnk4QWhHbVVldm53RFFNRXNUdzlPck53NC0yNXRyVSJ9",
  "sub": "did:jwk:eyJhbGciOiJFZERTQSIsImNydiI6IkVkMjU1MTkiLCJrdHkiOiJPS1AiLCJ4IjoiRE5NdDhrYUNPanlSVnk4QWhHbVVldm53RFFNRXNUdzlPck53NC0yNXRyVSJ9",
  "exp": 1715370592,
  "nbf": 1715368792,
  "jti": "urn:vc:uuid:1302309c-0729-4a00-8f40-93df77d41889"
}
```

Notice how every field in the vc is missing like id and issuer, these are populated from the claims like jti, iss
We actually dont support "empty" and only support Default like empty string and 0, so we will need to fix this issuer tracked here https://github.com/TBD54566975/web5-rs/issues/202


crates/credentials/src/vc.rs
```diff
@@ -6,16 +6,52 @@ use jwt::{
 };
 use serde::{Deserialize, Serialize};
 use std::{collections::HashMap, sync::Arc};
+use chrono::Utc;
```
KendallWeihe, line 9: this made me think, we should come to a consensus on this https://github.com/TBD54566975/web5-rs/issues/183

what you have here is fine, not proposing a change, just raising awareness

crates/credentials/src/vc.rs
```diff
@@ -6,16 +6,52 @@ use jwt::{
...
 #[derive(thiserror::Error, Debug)]
 pub enum CredentialError {
+    // JWT specific errors
     #[error(transparent)]
     JwtError(#[from] JwtError),
     #[error(transparent)]
     JwsError(#[from] JwsError),
+
+    // Credential-specific validation errors
+    #[error("Expiration date mismatch")]
```
KendallWeihe, line 23: embrace debate: I think we should try an be consistent with error strings and use all lower case characters

so the proposal here would be 

```suggestion
    #[error("expiration date mismatch")]
```

I've come to this opinion b/c that's the idiomatic approach in golang, and it exists for good reason, here are the reasons chad came up with:

- Error Wrapping: Lowercase error strings are easier to wrap in additional context without disrupting the flow of the text. This means you can prepend or append additional information to an existing error without resulting in awkward capitalization when errors are concatenated.
- Standard Library Consistency: It matches the style used by the standard library in Go, which generally begins error messages with lowercase.
- Error Chain Readability: It enhances the readability of error chains that might be formed by wrapping multiple errors together.

KendallWeihe: something like `ID` is different because it's an abbreviation so it would be capitalized

KendallWeihe: meh idk actually... I don't really care lol 😅 bigger 🐟 to fry

crates/credentials/src/vc.rs
```diff
@@ -6,16 +6,52 @@ use jwt::{
...
+    IdMismatch,
+    #[error("Subject mismatch")]
+    SubjectMismatch,
+    #[error("Missing issuer")]
+    MissingIssuer,
+    #[error("Misconfigured expiration date: {0}")]
+    MisconfiguredExpirationDate(String),
+
+    // Data model validation error
+    #[error("Validation error: {0}")]
```
KendallWeihe, line 41: FYI the reason this is failing is because you need to add the `thiserror::Error` to your `CredentialDataModelValidationError` enum below, and then attribute each enum value with the `#[error()]` enum

I just opened a ticket relevant to this, again not proposing a change with your code, just raising awareness https://github.com/TBD54566975/web5-rs/issues/184

KendallWeihe: also I would recommend this instead, because then you don't have to make the call to `.map_err` in your validate call below 

```rust
    #[error(transparent)]
    CredentialDataModelValidationError(#[from] CredentialDataModelValidationError),
```

crates/credentials/src/vc.rs
```diff
@@ -106,6 +198,42 @@ impl VerifiableCredential {
     }
 }
 
+pub fn validate_vc_data_model(vc: &VerifiableCredential) -> Result<(), CredentialDataModelValidationError> {
```
KendallWeihe, line 201: I'm generally more inclined to place this within the `verify()` function since that's the only place it's used, unless you know of a use case where this would need to be called from elsewhere? I know it'll make that function large (in number-of-lines), but that's okay, it's not any less large by splitting it out into a callable function. You could imagine, we expose this function as a callable function for consumers of our SDK and now we're locked in to always supporting it without inducing a breaking change (guiding principle being: be conservative in what is made public)

Open to being wrong on this

crates/credentials/src/vc.rs
```diff
@@ -106,6 +198,42 @@ impl VerifiableCredential {
...
+
+    // Required fields
+    if vc.context.is_empty() {
+        return Err(CredentialDataModelValidationError::MissingContext);
+    }
+
+    if vc.id.trim().is_empty() {
```
KendallWeihe, line 208: hmmm do we need this `trim()` call?

crates/credentials/src/vc.rs
```diff
@@ -90,13 +126,69 @@ impl VerifiableCredential {
...
+        validate_vc_data_model(&vc).map_err(CredentialError::ValidationError)?;
+
+        // exp MUST represent the expirationDate property, encoded as a UNIX timestamp (NumericDate).
+        if let Some(exp) = jwt_decoded.claims.registered_claims.expiration {
+            if let Some(expiration_date) = vc.expiration_date {
+                if exp != expiration_date {
+                    return Err(CredentialError::ExpirationDateMismatch);
+                }
+            } else {
+                vc.expiration_date = Some(exp);
```
KendallWeihe, line 143: Help me understand why we would want to set values on the `vc` here; I'm imagining we're doing this in other SDKs so I'm ignorant to the reasoning.

The reason I raise it is, it's a slightly weird developer experience to me, for a caller to pass in a JWT-encoded representation of a VC which has a set of properties A and then be handed the decoded VC with a set of properties B (that is, a different set of properties than what's in the JWT-encoded representation)

nitro-neal: image

the "official official" spec says that if exp is there you don't have to have vc.exirationDate.

To be explicit we want both, but to be compliant to spec we have a case if exp is there but there is no vc.exirationDate then we set it to exp 

but if they are out of sync in any other way then it fails

nitro-neal: its only like this for optional fields so its a slightly different truth table for required fields (basically it should have both and they should match)

I'll go through and double check I'm in sync with these truth tables

crates/credentials/src/vc.rs
KendallWeihe: at a minimum, I would recommend removing the `pub` and keep the function internal to the `vc` rust module

KendallWeihe: ya know I actually don't have strong opinions on this, apologies for the digress 

but I do think it's a good idea to remove the `pub` unless you're aware of a use case where this would be called external to this module

crates/credentials/src/vc.rs
nitro-neal: ^ check new pr message explaining this a bit better

KendallWeihe: oh interesting this is at the spec level! okay great well done 👏

crates/credentials/src/vc.rs
```diff
@@ -6,6 +6,8 @@ use jwt::{
     {Claims, JwtError, RegisteredClaims},
 };
 use serde::{Deserialize, Serialize};
+use std::cmp::PartialEq;
```
KendallWeihe, line 9: fun fact, you don't have to `use` this because it's a part of the "rust prelude" which means it's automatically imported

crates/credentials/src/vc.rs
```diff
@@ -127,25 +195,163 @@ impl VerifiableCredential {
         Ok(jwt)
     }
 
-    pub async fn verify(jwt: &str) -> Result<Self, CredentialError> {
+    pub async fn verify(jwt: &str) -> Result<Arc<VerifiableCredential>, CredentialError> {
```
KendallWeihe, line 198: I'd advise against introducing an `Arc` until it's absolutely necessary, it may lead to unnecessary complexity

crates/credentials/src/vc.rs
```diff
@@ -36,6 +95,15 @@ pub enum Issuer {
...
 }
 
+impl Issuer {
+    pub fn id(&self) -> &str {
+        match self {
+            Issuer::String(s) => s,
+            Issuer::Object(obj) => &obj.id,
+        }
+    }
+}
```
KendallWeihe, line 105: you can actually already do this by calling `.to_string()` on an instance of `issuer`, we do this in a few places already for example go look inside the `fn create_vc(issuer: Issuer) -> VerifiableCredential` function inside the tests module

KendallWeihe: the `impl Display for Issuer` implementation of the `fmt` method implements this functionality

crates/credentials/src/vc.rs
```diff
@@ -127,25 +195,163 @@ impl VerifiableCredential {
...
+        }
+        if registered_claims.jti.is_none() {
+            return Err(CredentialError::MissingJti);
+        }
 
-        Ok(jwt_decoded.claims.vc)
+        // TODO: Change this to None after this issue is resolved - https://github.com/TBD54566975/web5-rs/issues/202
+        // check claims match expected values
+        if vc.id == String::default() {
+            vc.id = registered_claims.jti.clone().unwrap();
```
KendallWeihe, line 220: so technically `unwrap` should be safe here since we're checking `if registered_claims.jti.is_none()` above, but I think it's ill advised to ever use `unwrap` or `expect` (or a number of other methods) in non-test code because there is the potential for a panic (any invocation which could panic should be avoided). instead, you could implement this one of a few ways, my recommended approach would be to remove the `if registered_claims.jti.is_none()` block above and handle that inline here like so:

```suggestion
            vc.id = registered_claims.jti.ok_or(CredentialError::MissingJti)?;
```

the same would hold true for all of the surrounding code which follows the same pattern

there's also a way to implement it using a `match` block

crates/credentials/src/vc.rs
```diff
@@ -127,25 +195,163 @@ impl VerifiableCredential {
...
+                } else if id != iss {
+                    return Err(CredentialError::IssuerMismatch);
+                }
+            }
+            Issuer::Object(named) => {
+                if &named.id != iss {
+                    return Err(CredentialError::IssuerMismatch);
+                }
+            }
+        }
```
KendallWeihe, line 258: Nothing wrong with yours but I find the `&mut` to be a bit esoteric to follow, this may be simpler (once again you could remove the `if registered_claims.issuer.is_none()` condition above b/c that occurs here inline)

```suggestion
        let iss = registered_claims
            .issuer
            .ok_or(CredentialError::MissingIss)?;
        let vc_issuer = vc.issuer.to_string();
        if vc_issuer.is_empty() {
            vc.issuer = Issuer::String(iss.clone());
        } else if iss != vc_issuer {
            return Err(CredentialError::IssuerMismatch);
        }
```

crates/credentials/src/vc.rs
```diff
@@ -127,25 +195,163 @@ impl VerifiableCredential {
...
+            if vc_exp == i64::default() {
+                vc.expiration_date = Option::from(exp);
+            } else if exp != vc_exp {
+                return Err(CredentialError::ExpirationDateMismatch);
+            }
+
+            let now = SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .unwrap()
+                .as_secs() as i64;
```
KendallWeihe, line 271: again I'd advise we stay away from calling `unwrap` in non-test code for the sake of error handling. also we should avoid using `as` casting as there's a baked in assumption there, instead we can use `i64::try_from()`

```suggestion
            let now = i64::try_from(SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs())?;
```

And then you'll have to add new error enums (and add the relevant `use` statements)

```rust
#[derive(thiserror::Error, Debug)]
pub enum CredentialError {
    #[error(transparent)]
    SystemTimeError(#[from] SystemTimeError),
    #[error(transparent)]
    TryFromIntError(#[from] TryFromIntError),
    // ...
```

crates/credentials/src/vc.rs
```diff
@@ -127,25 +195,163 @@ impl VerifiableCredential {
...
-
         Ok(jwt_decoded.claims.vc)
     }
 }
 
-// todo we should remove this altogether in the follow-up PR, but it would break bindings so leaving it for now
-pub async fn verify_vcjwt(jwt: &str) -> Result<Arc<VerifiableCredential>, CredentialError> {
-    let jwt_decoded = Jwt::verify::<VcJwtClaims>(jwt).await?;
-    Ok(Arc::new(jwt_decoded.claims.vc))
+fn validate_vc_data_model(
```
KendallWeihe, line 320: I feel like this should be an instance method on the `impl VerifiableCredential` implementation, so maybe like 

```rust
impl VerifiableCredential {
    // ...
    pub fn validate(&self) -> Result<(), CredentialDataModelValidationError> {
        // ...
    }
}
```

Which then brings into question, perhaps we should add `*_vcjwt` to the `sign()`, `verify()` and `decode()` functions. So that would be `sign_vcjwt()`, `verify_vcjwt()` and `decode_vcjwt()`. Rust doesn't support function overloading FWIW.

crates/credentials/src/vc.rs
```diff
@@ -127,25 +195,163 @@ impl VerifiableCredential {
...
+    vc: &VerifiableCredential,
+) -> Result<(), CredentialDataModelValidationError> {
+    // Required fields
+    if vc.id.is_empty() {
+        return Err(CredentialDataModelValidationError::MissingId);
+    }
+
+    if vc.context.first().map_or(true, |v| v != BASE_CONTEXT) {
```
KendallWeihe, line 328: maybe a bit easier on the eyes? I'll defer to your judgement (also holds true for `r#type` below)

```suggestion
    if vc.context.len() == 0 || vc.context[0] != BASE_CONTEXT {
```

crates/credentials/src/vc.rs
```diff
@@ -127,25 +195,163 @@ impl VerifiableCredential {
...
+        return Err(CredentialDataModelValidationError::MissingContext);
+    }
+
+    if vc.r#type.first().map_or(true, |v| v != BASE_TYPE) {
+        return Err(CredentialDataModelValidationError::MissingType);
+    }
+
+    if vc.issuer.id().is_empty() {
```
KendallWeihe, line 336: so yeah this would be an instance where you could do this instead

```suggestion
    if vc.issuer.to_string().is_empty() {
```

KendallWeihe:
this is really great work! I appreciate how you created follow-up ticket items. I've added a handful of comments probably worth addressing

crates/credentials/src/vc.rs
nitro-neal: ok happy to turn it off, was just sticking to the same signature it had before

crates/credentials/src/vc.rs
nitro-neal: ok awesome, thanks for this lol I was pulling my hair out

crates/credentials/src/vc.rs
nitro-neal: Well we've been down this path before

The thought was that the end user would get confused for
vc.validate()
vc.verify()

so the thought was to have a `verifyOptions` object, which you can have to only do validatePayload vs check specific verification vs all:
https://github.com/TBD54566975/web5-js/issues/425

crates/credentials/src/vc.rs
```diff
@@ -127,25 +195,163 @@ impl VerifiableCredential {
...
+        return Err(CredentialDataModelValidationError::MissingIssuer);
+    }
+
+    if vc.issuance_date.is_negative() {
+        return Err(CredentialDataModelValidationError::InvalidIssuanceDate);
+    }
+
+    // Validate expiration date if it exists
+    if let Some(expiration_date) = vc.expiration_date {
+        if expiration_date.is_negative() {
```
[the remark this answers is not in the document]
nitro-neal: Doh definitely. Adding now 💻

crates/credentials/src/vc.rs
nitro-neal: Lol lets use chrono :

`let now = Utc::now().timestamp();`

crates/credentials/src/vc.rs
nitro-neal: much better

crates/credentials/src/vc.rs
nitro-neal: even better `is_empty()` :)

crates/credentials/src/vc.rs
KendallWeihe: nice!

crates/credentials/src/vc.rs
KendallWeihe: okay cool, I may make a follow up GitHub Issue for this matter -- specifically, for the endeavor of maximizing modularity here at the rust core layer I think it's fitting to offer every single one of these little verification/validation checks as its own function. my thought process is a barbell design, wherein `verify()` does everything, but then we expose public facing methods for each individual one, so if developers want to create their own set of verification rules then they string the atomic ones together

KendallWeihe (approved):
🏆

Issue #39, opened 2023-12-05
Add `did:web` resolution test case for did uri that contains subdirectory


amika-sq:
Subdirectories may be included in a `did:web` uri, delimited with colons rather than slashes:
https://w3c-ccg.github.io/did-method-web/#method-specific-identifier

When initially implementing `did:web` support, we wanted to add a test for this with `did:web:users.tbddev.org:demo`, but at the time it included a `publicKeyHex` value and didn't refer to the proper context that has that defined.
[More info here.](https://github.com/TBD54566975/web5-rs/pull/24#discussion_r1414607470)

Making a task to go back and get a test for this written. We can create a new resolved document [here](https://github.com/tbdeng/tbd-didweb-internal/blob/main/chart/dids/demo) to test with!

PR #190, opened 2024-05-09
feat: add internal did:web resolution

### LINKED ISSUE #42: Implement our own `did:web` resolution
Similar to #41, there was feedback on https://github.com/TBD54566975/web5-rs/pull/24 that we should implement our own version of `did:web` resolution.

This task is to remove Spruce's `did-web` dependency from our `dids` crate and implement it ourselves.

enmand:
Remove dependency on the Spruce crates for did:web resolution, and implement `did:web` resolution internally in the `dids` crate.

Adds a test related to #39, but needs a trusted host for a real-resolution test for a path-based DID (does TBD have one already?)

Resolves #42.

_Note_: [e830f1d](https://github.com/TBD54566975/web5-rs/pull/190/commits/e830f1dd81bdd9e9e5f4317c38b2a409977d5d47) makes the `VerificationMessage` optional to match the current set of properties in [DID-Core](https://www.w3.org/TR/did-core/#core-properties) -- the [web5-spec](https://github.com/TBD54566975/web5-spec/blob/main/spec/did.md#did-document-data-model) disagrees with this, and I'm not sure who's right -- but the example hosted DID (https://www.tbd.website/.well-known/did.json) doesn't include a `VerificationMap`. The current implementation maps this to an empty `Vec<VerificationMethod>`, this PR changes that to an `Option<Vec<VerificationMap>>`.

### DIFF
--- a/crates/web5/src/dids/document.rs
+++ b/crates/web5/src/dids/document.rs
@@ -10,7 +10,7 @@ pub struct Document {
     pub controller: Option<Vec<String>>,
     #[serde(rename = "alsoKnownAs", skip_serializing_if = "Option::is_none")]
     pub also_known_as: Option<Vec<String>>,
-    #[serde(rename = "verificationMethod")]
+    #[serde(rename = "verificationMethod", default = "Vec::new")]
     pub verification_method: Vec<VerificationMethod>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub authentication: Option<Vec<String>>,
--- a/crates/web5/src/dids/methods/web/resolver.rs
+++ b/crates/web5/src/dids/methods/web/resolver.rs
@@ -39,46 +39,49 @@ impl Resolver {
             did_url: format!("https://{}/did.json", did_url),
         }
     }
-}
 
-// This trait implements the actual logic for resolving a DID URI to a DID Document.
-impl IntoFuture for Resolver {
-    type Output = Result<ResolutionResult, ResolutionError>;
-    type IntoFuture = Pin<Box<dyn Future<Output = Self::Output> + Send + Sync>>;
-
-    fn into_future(self) -> Self::IntoFuture {
+    async fn resolve(url: String) -> Result<ResolutionResult, ResolutionError> {
         let mut headers = HeaderMap::new();
         headers.append(
             reqwest::header::USER_AGENT,
             HeaderValue::from_static(USER_AGENT),
         );
 
-        Box::pin(async move {
-            let client = reqwest::Client::builder()
-                .default_headers(headers)
-                .build()
-                .map_err(|_| ResolutionError::InternalError)?;
+        let client = reqwest::Client::builder()
+            .default_headers(headers)
+            .build()
+            .map_err(|_| ResolutionError::InternalError)?;
 
-            let response = client
-                .get(&self.did_url)
-                .send()
+        let response = client
+            .get(&url)
+            .send()
+            .await
+            .map_err(|_| ResolutionError::InternalError)?;
+
+        if response.status().is_success() {
+            let did_document = response
+                .json::<Document>()
                 .await
-                .map_err(|_| ResolutionError::InternalError)?;
+                .map_err(|_| ResolutionError::RepresentationNotSupported)?;
 
-            if response.status().is_success() {
-                let did_document = response
-                    .json::<Document>()
-                    .await
-                    .map_err(|_| ResolutionError::RepresentationNotSupported)?;
+            Ok(ResolutionResult {
+                did_document: Some(did_document),
+                ..Default::default()
+            })
+        } else {
+            Err(ResolutionError::NotFound)
+        }
+    }
+}
 
-                Ok(ResolutionResult {
-                    did_document: Some(did_document),
-                    ..Default::default()
-                })
-            } else {
-                Err(ResolutionError::NotFound)
-            }
-        })
+// This trait implements the actual logic for resolving a DID URI to a DID Document.
+impl IntoFuture for Resolver {
+    type Output = Result<ResolutionResult, ResolutionError>;
+    type IntoFuture = Pin<Box<dyn Future<Output = Self::Output>>>;
+
+    fn into_future(self) -> Self::IntoFuture {
+        let did_url = self.did_url;
+        Box::pin(async move { Self::resolve(did_url).await })
     }
 }
 
--- a/crates/web5/src/dids/methods/web/resolver.rs
+++ b/crates/web5/src/dids/methods/web/resolver.rs
@@ -16,7 +16,7 @@ const PORT_SEP: &str = "%3A";
 
 const USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"));
 
-/// Resolver is the implementation of the did:web method for resolcing DID URIs. It is responsible
+/// Resolver is the implementation of the did:web method for resolving DID URIs. It is responsible
 /// for fetching the DID Document from the web according for the did-web spec.
 pub struct Resolver {
     did_url: String,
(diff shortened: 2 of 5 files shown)


crates/dids/src/method/web/resolver.rs
```diff
@@ -0,0 +1,99 @@
...
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[tokio::test]
+    async fn resolution_success() {
```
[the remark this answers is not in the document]
enmand: Yep, I can certainly do that!

crates/dids/src/document.rs
```diff
@@ -10,8 +10,8 @@ pub struct Document {
     pub controller: Option<Vec<String>>,
     #[serde(rename = "alsoKnownAs", skip_serializing_if = "Option::is_none")]
     pub also_known_as: Option<Vec<String>>,
-    #[serde(rename = "verificationMethod")]
-    pub verification_method: Vec<VerificationMethod>,
+    #[serde(rename = "verificationMethod", skip_serializing_if = "Option::is_none")]
+    pub verification_method: Option<Vec<VerificationMethod>>,
```
enmand, line 14: How do you all feel about this change? 6a54d270387965d396529f1311174b2a1a22fdd8 has some reasoning -- did-core and the web5 spec disagree so I'm not sure who's right. This example DID document on https://tbd.website/.well-known/did.json doesn't have a `VerificationMethod` which is what sparked this change.

Maybe it's worth a separate PR for discussion aside from the `did:web` resolution PR?

KendallWeihe:
I see review has been requested, nice! This may sit for a bit as I'll be AFK for the next week, but others may get to it in the mean time. Nevertheless, we'll get this reviewed sooner or later! Thanks 🙌

crates/dids/src/document.rs
decentralgabe: yes, though I am wondering whether this guidance should still apply to the rust code.
originally it was to support a consistent feature set across languages that handled optional/polyglot type properties consistenly
however, if rust is able to directly support the did core spec then maybe our guidance is superfluous

suffice to say - in this case I would be fine to follow DID core and ignore our spec - but open to counter points

enmand:
Looks like this PR is out-of-date with the recent crate changes. I should have time later this week to rebase it.

enmand:
Rebased on the changes for the crate re-organization. Let me know if you want me to separate the `VerificationMethod` stuff into a separate PR.

crates/web5/src/dids/document.rs
```diff
@@ -10,8 +10,8 @@ pub struct Document {
     pub controller: Option<Vec<String>>,
     #[serde(rename = "alsoKnownAs", skip_serializing_if = "Option::is_none")]
     pub also_known_as: Option<Vec<String>>,
-    #[serde(rename = "verificationMethod")]
-    pub verification_method: Vec<VerificationMethod>,
+    #[serde(rename = "verificationMethod", skip_serializing_if = "Option::is_none")]
+    pub verification_method: Option<Vec<VerificationMethod>>,
```
KendallWeihe, line 14: What was the motivating factor for moving this property to optional?

We currently require at least 1 verification method via [our web5-spec](https://github.com/TBD54566975/web5-spec/blob/main/spec/did.md#did-document-data-model), albeit this is not to be considered entirely orthodox and open to changing.

KendallWeihe: Oops, I can see this has already been discussed in a prior comment. Apologies. Disregard this comment.

crates/dids/src/document.rs
KendallWeihe: Okay cool this is a great matter to work through. I'm in favor of enabling support for the DID Core spec over our `web5-spec` here in Rust. However, I propose we don't implement that change here, because there are more implications which need attending to. Primarily, the ones which come to mind are:

1. did:jwk requires exactly one verification method, and so we should add error cases
2. did:dht requires at least one verification method, and so we should add error cases
3. we'll need to also account for this change in the [APID](https://github.com/TBD54566975/web5-rs/pull/229) and consider the downstream implications for Kotlin & Swift target languages

@enmand could we parse these changes out of this PR, and then I stubbed in #231 for subsequent work

enmand: Sounds good! I've opened https://github.com/TBD54566975/web5-rs/pull/233 with the changes to enable optional verification methods and rebased it out of this PR.

crates/web5/src/dids/methods/web/resolver.rs
```diff
@@ -0,0 +1,116 @@
...
+    identifier::Identifier,
+    resolver::{ResolutionError, ResolutionResult},
+};
+
+// PORT_SEP is the : character that separates the domain from the port in a URI.
+const PORT_SEP: &str = "%3A";
+
+const USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"));
+
+/// Resolver is the implementation of the did:web method for resolcing DID URIs. It is responsible
```
KendallWeihe, line 19:
```suggestion
/// Resolver is the implementation of the did:web method for resolving DID URIs. It is responsible
```

crates/web5/src/dids/document.rs
```diff
@@ -10,7 +10,7 @@ pub struct Document {
     pub controller: Option<Vec<String>>,
     #[serde(rename = "alsoKnownAs", skip_serializing_if = "Option::is_none")]
     pub also_known_as: Option<Vec<String>>,
-    #[serde(rename = "verificationMethod")]
+    #[serde(rename = "verificationMethod", default = "Vec::new")]
```
KendallWeihe, line 13: Clever. This may have implications but I'm fine with moving forward with it.

crates/web5/src/dids/methods/web/resolver.rs
```diff
@@ -0,0 +1,116 @@
...
+use crate::dids::{
+    document::Document,
+    identifier::Identifier,
+    resolver::{ResolutionError, ResolutionResult},
+};
+
+// PORT_SEP is the : character that separates the domain from the port in a URI.
+const PORT_SEP: &str = "%3A";
+
+const USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"));
```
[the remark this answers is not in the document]
enmand: Mostly because it's a way for the did:web server to understand what versions/libraries it's dealing with, though given it's not a service-to-service environment and there are many callers maybe it's not as important. I'm not strong on keeping it or removing it -- I only added it because it was a natural thing for me to include, but also happy to not have it included.

Re. WASM environments -- that's a good questions. I'll implement it in a browser/WASM environment soon for the DWN project I'm working on, and if this is included I can report back (otherwise, I'll probably have other places where I make `reqwest` requests in the DWN so I can try it there). One challenge with WASM in general is networking, but WASIX and other modern preview specs include some networking (including HTTP stubbed by `fetch`, iirc)

crates/web5/src/dids/document.rs
enmand: I added it because the example `did:web` (https://tbd.website/.well-known/did.json) was failing to Deserialize, complaining about `verification_methods` being missing -- though maybe there's another way (aside from #233)?

decentralgabe: We should update that DID because it's not really adding any value. I'm not sure how it can be used in its current state.

KendallWeihe (approved):
Really fantastic work ✅

crates/web5/src/dids/methods/web/resolver.rs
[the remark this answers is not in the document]
KendallWeihe: I'm wrangling around code into [the APID](https://github.com/TBD54566975/web5-rs/pull/229) in [this PR](https://github.com/TBD54566975/web5-rs/pull/234). I'm going to go ahead and merge this code and copy everything over into the temporary `apid` module, but I'm going to remove this feature. It's probably something we want, but I want to approach these matters conservatively because of downstream impacts to cross-language binding. I appreciate the thought! And I suspect we'll add it back at some point once we fully comprehend the implications.

PR #192, opened 2024-05-09
Add Presentation Definition and use test vectors

diehuxx:
Copy algorithm for `selectCredentials` laid out here https://github.com/TBD54566975/web5-go/pull/143.
Use web5-spec submodule to get test vectors. As we use test vectors more, we'll want to establish a common pattern and DRY up the code.


crates/credentials/src/pex/presentation_definition.rs
```diff
@@ -0,0 +1,292 @@
...
+    pub r#type: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub pattern: Option<String>,
+    #[serde(rename = "const", skip_serializing_if = "Option::is_none")]
+    pub const_value: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub contains: Option<Box<Filter>>,
+}
+
+fn generate_token() -> String {
```
nitro-neal, line 88: why not uuid? 

:caution: chad generated:
```
[dependencies]
uuid = { version = "1.0", features = ["v4"] } 

extern crate uuid;
use uuid::Uuid;

fn generate_uuid_token() -> String {
    let my_uuid = Uuid::new_v4();
    my_uuid.to_string()
}
```

diehuxx: Good call

crates/credentials/src/pex/presentation_definition.rs
```diff
@@ -0,0 +1,292 @@
...
+    }
+}
+
+struct TokenizedField<'a> {
+    pub token: String,
+    pub path: &'a String,
+}
+
+impl InputDescriptor {
+    pub fn select_credentials(&self, vc_jwts: &Vec<String>) -> Result<Vec<String>> {
```
nitro-neal, line 113: lol pex is so hard, can you give some documentation on the overall algorithm here?

crates/credentials/src/pex/presentation_definition.rs
```diff
@@ -0,0 +1,292 @@
...
+        Ok(matches.into_iter().collect())
+    }
+}
+
+struct TokenizedField<'a> {
+    pub token: String,
+    pub path: &'a String,
+}
+
+impl InputDescriptor {
```
nitro-neal, line 112: in the other libs we have 

PresentationExchange.select_credentials
PresentationExchange.createPresentationFromCredentials

...

Was the plan to keep it similar or are there other api design changes in the pipeline?

https://github.com/TBD54566975/web5-js/tree/main/packages/credentials#presentationexchange

nitro-neal (approved):
If it works it works! My understanding of the rust language is still `rusty` 😃 
but this works by 

1. doing some preprocessing creating a token mapping for each field
2. you do processing on each element in the jwt payload and see if it matches?
3. then do the jsonpath and if it matches then it is successful

something like that

crates/credentials/src/pex/presentation_definition.rs
diehuxx: The priority for now (keep me honest @frankhinek @mistermoe) is only `select_credentials` since that is needed for TBDex but `createPresentationFromCredentials` is not. Instead of using Presentation Submissions in TBDex RFQ messages, we submit a list of claims in the `Rfq.data.required_claims` field.

crates/credentials/src/pex/presentation_definition.rs
diehuxx: For sure. I'll also clean up the code a bit. I'm not good at making things readable in rust yet

mistermoe:
@nitro-neal 

the algorithm compiles an `input_descriptor`'s `constraints.fields` into:
* a list of tokenized fields (1 path, 1 token)
* an aggregate JSON schema includes the `filter` of each field element that has a `filter`. The token is used as the name of the property in the json schema

once an ID is compiled into ^, the tokenized fields are iterated through to build a `selectionCandidate` object that is a map of `token: <value found at path>`. This selection candidate is then eval'd against the aggregate json schema

crates/credentials/src/pex/presentation_definition.rs
diehuxx: Done. If it's still hard to read, let me know and I'll take another pass :)

KendallWeihe:
this is really amazing work 💯 

created #195 and #196 as follow-ups

PR #199, opened 2024-05-11
Remove unnecessary Arc<T> type

### LINKED ISSUE #198: Remove unnecessary `Arc<T>` type for `Key` trait's `jwk()` method
```rust
pub trait Key: Send + Sync {
    fn alias(&self) -> Result<String, KeyError>;
    fn jwk(&self) -> Result<Arc<Jwk>, KeyError>;
}
```

There's no good reason for the `jwk()` method in the `Key` trait to not return a `Jwk`; the `Arc<T>` is an artifact of some prior workshopping with UniFFI bindings. UniFFI types implemented as interfaces and then user in function signatures are required to be `Arc`. This may arise again once we start building out bindings, but until that point it's unnecessary overhead (case in point, [this comment](https://github.com/TBD54566975/web5-rs/pull/156#discussion_r1595731785))

KendallWeihe:
Closes #198

### DIFF
--- a/crates/dids/src/method/jwk.rs
+++ b/crates/dids/src/method/jwk.rs
@@ -36,7 +36,7 @@ impl Create<DidJwkCreateOptions> for DidJwk {
         let key_alias = key_manager.generate_private_key(options.curve, Some("0".to_string()))?;
         let public_key = key_manager.get_public_key(&key_alias)?;
         let public_jwk = public_key.jwk()?;
-        let jwk_string = serde_json::to_string(public_jwk.as_ref()).map_err(|_| {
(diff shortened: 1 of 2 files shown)


crates/jws/src/lib.rs
```diff
@@ -479,7 +479,8 @@ mod tests {
             &header,
             &payload,
         )
-        .unwrap() + "123";
+        .unwrap()
```
KendallWeihe, line 482: interesting, added https://github.com/TBD54566975/web5-rs/issues/200 to try and avoid this in the future

diehuxx (approved):

PR #201, opened 2024-05-12
Util to load test vector

### LINKED ISSUE #196: Create shared/reusable code for loading test vectors
Test vectors first integrated and used here https://github.com/TBD54566975/web5-rs/pull/192 (also added git submodule to just recipe in #195)

We'll need to use test vectors across many places in the repo, so we should consider creating a single places for loading the test fixtures. It may not make sense here, but at the least let's rule it out.

We did something like this over in `web5-go` https://github.com/TBD54566975/web5-go/blob/main/testvectors.go

diehuxx:
Closes #196

### DIFF
--- a/crates/credentials/src/presentation_definition.rs
+++ b/crates/credentials/src/presentation_definition.rs
@@ -236,12 +236,13 @@ impl JsonSchemaBuilder {
 #[cfg(test)]
 mod tests {
     use std::collections::HashSet;
-    use std::fs;
+
+    use test_helpers::TestVectorFile;
 
     use super::PresentationDefinition;
 
     #[derive(Debug, serde::Deserialize)]
-    struct SelectCredentialsVectorInput {
+    struct VectorInput {
         #[serde(rename = "presentationDefinition")]
         pub presentation_definition: PresentationDefinition,
         #[serde(rename = "credentialJwts")]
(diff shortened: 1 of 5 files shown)


crates/credentials/src/presentation_definition.rs
```diff
@@ -236,47 +236,30 @@ impl JsonSchemaBuilder {
...
-    }
-
     #[test]
     fn test_web5_spec_test_vectors() {
-        let json_path =
-            "../../web5-spec/test-vectors/presentation_exchange/select_credentials.json";
-        let vectors = load_json_fixture(json_path);
+        let path = "presentation_exchange/select_credentials.json";
+        let vectors: TestVectorFile<VectorInput, VectorOutput> =
+            TestVectorFile::load_from_path(path);
```
KendallWeihe, line 262: the typing of the generics is also possible like this:

```suggestion
        let vectors = TestVectorFile::<VectorInput, VectorOutput>::load_from_path(path);
```

both are semantically equivalent, but the type definition for the generics live in different places syntactically

I tend to prefer the proposed way but it's potato, potahto

KendallWeihe (approved):
this is eggsellent, love how we are able to to standardize the `TestVector` and `TestVectorFile` type structs 

I do wonder if it would be wise to disambiguate `crates/` versus something like `internal-crates/` but I don't have any great ideas right now

crates/credentials/src/presentation_definition.rs
diehuxx: TIL -- I like the latter option too, thanks for showing me!

PR #203, opened 2024-05-13
Add ci step to fail given formatting is needed

### LINKED ISSUE #200: GitHub CI workflow should fail if `cargo fmt` edits source code
Prior to #187, if the call to `cargo fmt` actually changed the underlying source code then the pipeline would fail, which would indicate to the developer they need to format their code and then push changes. But since we moved to #187 that feature is no longer enforced, I noticed because in #199 I had formatting changes which were from where #193 didn't run `just lint` prior to merging, but it wasn't caught because the pipeline didn't fail. The pipeline shouldn't allow this to happen.

KendallWeihe:
Closes #200

### DIFF
--- a/crates/credentials/src/presentation_definition.rs
+++ b/crates/credentials/src/presentation_definition.rs
@@ -262,7 +262,7 @@ mod tests {
             TestVectorFile::load_from_path(path);
 
         for vector in vectors.vectors {
-            let presentation_definition = vector.input.presentation_definition;
+        let presentation_definition = vector.input.presentation_definition;
             let vc_jwts = vector.input.credential_jwts;
             let error_msg = format!(
                 "Selected Credential test vector ({}) should not have thrown error",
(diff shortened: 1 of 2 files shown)


crates/credentials/src/presentation_definition.rs
```diff
@@ -262,7 +262,7 @@ mod tests {
             TestVectorFile::load_from_path(path);
 
         for vector in vectors.vectors {
-            let presentation_definition = vector.input.presentation_definition;
+        let presentation_definition = vector.input.presentation_definition;
```
KendallWeihe, line 265: Testing if this works... will undo once pipeline proves failure

KendallWeihe: Alright and that's successfully passing now ✅  https://github.com/TBD54566975/web5-rs/actions/runs/9067119392/job/24911514344?pr=203

nitro-neal (approved):

Issue #158, opened 2024-05-06
Move import & export functions to new trait called `UnsafeKeyManager`


KendallWeihe:
Currently the `export_private_keys()` and `import_private_keys()` are a part of the `KeyManager` trait, but this is undesirable for two reasons. First, we want to promote the exclusion of the implementation of these methods because moving keys out of the given key store poses security risks. Second, we want to avoid implementations which return an error for "unsupported" or "not implemented" and instead shift-left such that the implementation is determinable at build time (throwing an error isn't discovered until runtime whereas the implementation of a rust trait is determinable at build time).

Instead, let's introduce a new trait, appropriately named `UnsafeKeyManager` which resolves both of these issues; it's named as such as to promote the exclusion of it (b/c what developer wants to use something "unsafe"? I suppose there is an assumption in there but I'm okay making it) & also the implementation of such a trait will be determinable at build time.

frankhinek:
## Alternative Proposal

Follow an approach similar to what has been implemented in JS, Kotlin, Dart, and Go:

1. Remove the `export_private_keys` and `import_private_keys` methods.

2. Add two additional traits:
- `KeyImporter`
- `KeyExporter`

These can optionally be implemented by concrete implementations of `KeyManager` when this functionality is desirable.

## Background

Not all `KeyManager` implementations can support exporting keys.  For example, this typically isn't possible with both mobile device and cloud-based HSMs.  Additionally, some key management systems can support importing of externally created key material, but this can be rather involved (e.g., see docs on AWS KMS and Google CloudKMS key import).

KendallWeihe:
It's a good point. I hadn't considered a use case wherein a key manager supports importing but not exporting (or vice versa). `KeyImporter` and `KeyExporter` doesn't convey the intent of moving keys around being an unsafe behavior, but I suppose it's fine. Developer beware. I'm in favor of the proposed alternative @frankhinek 👍

KendallWeihe:
Completed in #204 

Opened #211 for exporter

PR #204, opened 2024-05-13
Refactor KeyManager

diehuxx:
Partially addresses https://github.com/TBD54566975/web5-rs/issues/158

### Context
The `KeyManager` and `KeyStore` traits are fuzzy abstractions -- the conceptual boundary between a "store" and a "manager" is not clear to me. `KeyManager` currently seems to be a wrapper around `KeyStore` with little-to-no added functionality. Also, the traits themselves have methods `export_private_keys()` and `import_private_keys()` which ought not to be coupled to the notion of "management" which entails storing private keys, signing, and getting public keys. See https://github.com/TBD54566975/web5-rs/issues/158 for more.

### Design
1. Remove the `KeyStore` trait entirely. Move logic from struct `InMemoryKeyStore` into `LocalKeyManager`.
2. Pare down `KeyManager` to just three methods: `generate_private_key()`, `sign()`, and `get_public_key()`.
3. Add trait `KeyImporter` with one required method `import_with_alias()` and one provided method `import()` which generates its own alias.

This PR does NOT add the `KeyExporter` trait described in https://github.com/TBD54566975/web5-rs/issues/158 because I do not see an immediate use case in `web5-rs`. Exporting private keys is risky stuff. Until we have a clear need for a standard interface for exporting, we ought to omit it.

### Open questions
1. What is the purpose of the `.alias()` method in the `Key` trait? Is it a unique digest? Whatever it is, we should probably rename the method.
2. What is the expected behavior of `KeyImporter` when a key alias is overwritten with a new private key? Should that be an error? Is it up to each individual `impl` of `KeyImporter`? I lean towards the former.


crates/keys/src/key_manager/local_key_manager.rs
```diff
@@ -31,41 +38,81 @@ impl KeyManager for LocalKeyManager {
         curve: Curve,
         key_alias: Option<String>,
     ) -> Result<String, KeyManagerError> {
-        let key_alias = self.key_store.generate_new(curve, key_alias)?;
+        let private_key = Arc::new(match curve {
```
KendallWeihe, line 41: FYI #205 (not necessary for this PR, just raising awareness)

crates/keys/src/key_manager/local_key_manager.rs
```diff
@@ -31,41 +38,81 @@ impl KeyManager for LocalKeyManager {
...
+            Curve::Ed25519 => Ed25519::generate(),
+            Curve::Secp256k1 => Secp256k1::generate(),
+        }?);
+        let key_alias = match key_alias {
+            Some(key_alias) => key_alias,
+            None => private_key.compute_thumbprint()?,
+        };
```
KendallWeihe, line 48: Alternative, I don't care one way or the other 

```suggestion
        let key_alias = key_alias.ok_or_else(|| private_key.compute_thumbprint())?;
```

crates/keys/src/key_manager/mod.rs
```diff
@@ -39,23 +42,22 @@ pub trait KeyManager: Send + Sync {
...
+        private_key: Arc<dyn PrivateKey>,
+        key_alias: &str,
+    ) -> Result<(), KeyManagerError>;
+
+    /// Imports a private key into the key manager using private_key.alias() as the alias
+    /// Returns the key alias
+    fn import(&self, private_key: Arc<dyn PrivateKey>) -> Result<String, KeyManagerError> {
+        let key_alias = private_key.alias()?;
+        self.import_with_alias(private_key, &key_alias)?;
+        Ok(key_alias)
```
KendallWeihe, line 61: this is clever 👏

KendallWeihe (approved):

KendallWeihe:
Alright a handful of things...

- This is amazing work. I've ✅'d but I propose we let it sit for a couple days at least to give people time to chime in if they choose.
- I respect taking the discretion to remove the `KeyStore` and I approve. We have a guiding principal here in rust core to "_prefer modularity over convenience_" but in this case, we don't have a concrete use case (that we can even think of) where decoupling the abstractions is necessary, so yeah I think it's wise to remove.
- I appreciate the named functions `import_with_alias` and `import` rather than using a single function with `Option` param. I think this is the ideal pattern here in our rust initiative -- it's slightly more verbose than what we're accustomed to with other languages, but it'll pay dividends IMO as we start bindings things.
- 💯 on not implementing the `KeyExporter` trait. I think what I'll probably do is eventually close #158 and create a new ticket for the export functionality, which hopefully just sits there forever. I'll handle that.

> What is the purpose of the .alias() method in the Key trait? Is it a unique digest? Whatever it is, we should probably rename the method.

It's a good point... the way I think about it is, the key alias is akin to a primary key in a db, so it's self referential with respect to a key. It's slightly confusing though because we enable callers of key manager methods to override the key alias, so it's a matter which is conceptualized at the key manager level, but at the same time we implement the `alias` for the `Jwk` which is the thumbprint. 🤔 created https://github.com/TBD54566975/web5-rs/issues/206

> What is the expected behavior of KeyImporter when a key alias is overwritten with a new private key? Should that be an error? Is it up to each individual impl of KeyImporter? I lean towards the former.

Yeah that should be an error. So the requirement is: _key's cannot be imported which have a key alias which pre-exist in the key manager_. Created https://github.com/TBD54566975/web5-rs/issues/207 @diehuxx feel free to go ahead and do that here or not and we can follow up, your call.

crates/keys/src/key_manager/local_key_manager.rs
diehuxx: ty! I prefer the `match` statement here

mistermoe:
How is a portable did created without a key exporter?

KendallWeihe:
> How is a portable did created without a key exporter?

It can't. But, TBD. This is a hot topic, so perhaps I'll open a ticket to discuss the full merits. I know this has been discussed at length a while back, but it's appropriate to take a bit of time to reopen the matter with the rust initiative.

PR #209, opened 2024-05-14
Add bep44 encoded for did:dht

diehuxx:
This PR does not yet make use of the bep44 mod. In a follow up PR that implements `DidDht::create`, this encoder will actually be used.

See also the dart implementation: https://github.com/TBD54566975/web5-dart/blob/main/packages/web5/lib/src/dids/did_dht/bep44.dart

### DIFF
--- a/crates/dids/src/method/dht/bep44.rs
+++ b/crates/dids/src/method/dht/bep44.rs
@@ -7,16 +7,19 @@ use byteorder::{BigEndian, ReadBytesExt, WriteBytesExt};
 use crypto::CryptoError;
 use keys::key::{KeyError, PublicKey};
 
+/// Minimum size of a bep44 encoded message
+/// Signature is 64 bytes and seq is 8 byets
 const MIN_MESSAGE_LEN: usize = 72;
-const MAX_MESSAGE_LEN: usize = 1072;
+/// Maximum size of a bep44 v field
+const MAX_V_LEN: usize = 1000;
+/// Maximum size a bep44 encoded message
+const MAX_MESSAGE_LEN: usize = MAX_V_LEN + MIN_MESSAGE_LEN;
 
 /// Errors that can occur when working with Bep44 messages for did:dht.
 #[derive(thiserror::Error, Debug)]
 pub enum Bep44EncodingError {
     #[error(transparent)]
     SystemTimeError(#[from] SystemTimeError),
-    #[error("Could not coerce microseconds since epoch into u64")]
-    TimeError,
     #[error(transparent)]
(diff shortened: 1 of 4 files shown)


crates/dids/src/method/dht/bep44.rs
```diff
@@ -0,0 +1,233 @@
...
+    io::Cursor,
+    time::{SystemTime, SystemTimeError, UNIX_EPOCH},
+};
+
+use byteorder::{BigEndian, ReadBytesExt, WriteBytesExt};
+use crypto::CryptoError;
+use keys::key::{KeyError, PublicKey};
+
+const MIN_MESSAGE_LEN: usize = 72;
+const MAX_MESSAGE_LEN: usize = 1072;
```
decentralgabe, line 11: Might want to split this into something like "max V length" since the sign and seq are separate and we need to enforce that v is no longer than 1000b

crates/dids/src/method/dht/bep44.rs
```diff
@@ -0,0 +1,233 @@
...
+/// See [BEP44 Specification](https://www.bittorrent.org/beps/bep_0044.html)
+impl Bep44Message {
+    pub fn new<F>(message: &[u8], sign: F) -> Result<Self, Bep44EncodingError>
+    where
+        F: Fn(Vec<u8>) -> Result<Vec<u8>, KeyError>,
+    {
+        let seq: u64 = SystemTime::now()
+            .duration_since(UNIX_EPOCH)?
+            .as_micros()
+            .try_into()
```
decentralgabe, line 84: I'm not sure what this does but the spec has the seq in seconds https://did-dht.com/#create

decentralgabe (commented):
Nice work!

crates/dids/src/method/dht/bep44.rs
diehuxx: Ah that's a mistake because copied it blindly from web5-dart. Here's a PR to update it in dart https://github.com/TBD54566975/web5-dart/pull/87

KendallWeihe (approved):
good to go once that comment on seconds is addressed!

and yeah I guess also maybe we should enforce the 1000 byte limitation in the `new()` function

crates/dids/src/method/dht/bep44.rs
diehuxx: I added a const `MAX_V_LEN` and some documentation to keep it more clear. We need to do the check on `MIN_MESSAGE_LEN` at the beginner of `decode()` regardless so we don't try to access an index out of bounds.

PR #210, opened 2024-05-15
Rename 'dids' crate 'method' module to 'methods' (plural)

KendallWeihe:
This way we match `dids`, `credentials`, and `keys` -- all of which are plural.


decentralgabe (approved):

diehuxx (approved):
theres a methods to the madness

Issue #103, opened 2024-04-21
Add doc comments everywhere


KendallWeihe:
As a part of https://github.com/TBD54566975/web5-rs/pull/97/ I didn't include doc comments because it was already way past the point of being wieldy, we need doc comments everywhere. Break this work up into multiple PRs where appropriate.

KendallWeihe:
Justfile changes:

add `cargo install watchexec-cli` to the `setup` recipe, and then add a new recipe

```justfile
docs: setup
  cargo doc --no-deps --open
  watchexec -w crates/ "cargo doc --no-deps"
```

---

When we write doc comments, use syntax such that referenced types/traits/enums/functions are href clickable, for example, the below doc comment for `JwsHeader` ought to have href click links for the DID Document and KeySelector (with names adjusted as needed)


<img width="799" alt="Screenshot 2024-05-16 at 1 47 43 AM" src="https://github.com/TBD54566975/web5-rs/assets/3602993/fcfb2017-b29e-4a15-aa4f-139a21a5e590">

PR #213, opened 2024-05-16
Add key Signer trait in place of bearer_did+key_selector #208

### LINKED ISSUE #208: Replace `bearer_did`+`key_selector` as the mechanism for signing with a `Sign` type
Currently, in a number of places we follow a pattern such as:

```rust
    pub fn sign(
        &self,
        bearer_did: &BearerDid,
        key_selector: &KeySelector,
    ) -> Result<String, CredentialError> {
```

...for the the calling code to indicate what keys to sign with. Instead, we should replace this `bearer_did`+`key_selector` input parameter pattern with a generalized `Sign` function type. 

The `bearer_did`+`key_selector` idea is conceptualized as: "_here is my bearer DID, which includes a reference to my key manager, and here's the key I want to use to sign, now you handle the actual signing_."

This is beneficial in that it places DID's as first-class citizens in out developer experience, but it's also constricting in that we only enable cryptographic signing through the existence of a bearer DID. A guiding principal we embrace in the rust initiative is to _prefer modularity over convenience_. This is an instance where we have convenience but we need modularity. We will have use cases where we need to support cryptographic signing without the existence of a bearer DID.

It'll likely make the most sense to place this `Sign` function in the `crypto` package.

We need to ensure this won't cause our bindings initiative to increase in complexity by an order of magnitude. First things first is to prove how such a paradigm shift would be implemented with UniFFI bindings. UniFFI UDL supports a feature known as [callback interfaces](https://mozilla.github.io/uniffi-rs/udl/callback_interfaces.html), but it appears they may not be considered to be "soft deprecated."

> These are described in both UDL and proc-macros as an explicit "callback interface". They are (soft) deprecated, remain now for backwards compatibility, but probably should be avoided.

In which case, perhaps there is a different implementation solution for passing function types as function parameters.

KendallWeihe:
Closes #208 

Started to write doc comments but realized this was already a lot, so added a comment to the existing ticket for doc comments https://github.com/TBD54566975/web5-rs/issues/103#issuecomment-2114082892

### DIFF
--- a/crates/credentials/src/vc.rs
+++ b/crates/credentials/src/vc.rs
@@ -1,10 +1,10 @@
 use core::fmt;
-use dids::{bearer::BearerDid, document::KeySelector};
-use jws::JwsError;
+use jws::{JwsError, JwsHeader};
 use jwt::{
     jws::Jwt,
     {Claims, JwtError, RegisteredClaims},
 };
+use keys::key_manager::Signer;
 use serde::{Deserialize, Serialize};
 use std::{
     collections::HashMap,
@@ -107,8 +107,8 @@ impl VerifiableCredential {
 
     pub fn sign(
         &self,
-        bearer_did: &BearerDid,
-        key_selector: &KeySelector,
+        signer: Arc<dyn Signer>,
+        jws_header: JwsHeader,
     ) -> Result<String, CredentialError> {
         let claims = VcJwtClaims {
             registered_claims: RegisteredClaims {
@@ -122,7 +122,7 @@ impl VerifiableCredential {
             vc: self.clone(),
         };
 
-        let jwt = Jwt::sign(bearer_did, key_selector, None, &claims)?;
+        let jwt = Jwt::sign(signer, jws_header, &claims)?;
 
         Ok(jwt)
     }
@@ -162,7 +162,8 @@ mod test {
     use super::*;
     use crypto::Curve;
     use dids::{
-        document::VerificationMethodType,
+        bearer::BearerDid,
+        document::{KeyIdFragment, KeySelector, VerificationMethodType},
         methods::{
             jwk::{DidJwk, DidJwkCreateOptions},
             Create,
@@ -313,7 +314,14 @@ mod test {
         let key_selector = KeySelector::MethodType {
             verification_method_type: VerificationMethodType::VerificationMethod,
         };
-        let vcjwt = vc.sign(&bearer_did, &key_selector).unwrap();
+
+        let key_id = bearer_did.document.verification_method[0].id.clone();
+        let key_alias = KeyIdFragment(key_id.clone()).splice_key_alias();
+        let signer = bearer_did.key_manager.get_signer(&key_alias).unwrap();
+
+        let jws_header = JwsHeader::from_did_document(&bearer_did.document, &key_selector).unwrap();
+
+        let vcjwt = vc.sign(signer, jws_header).unwrap();
         assert!(!vcjwt.is_empty());
 
         let verified_vc = VerifiableCredential::verify(&vcjwt).await.unwrap();
--- a/bindings/uniffi/src/lib.rs
+++ b/bindings/uniffi/src/lib.rs
@@ -1,5 +1,27 @@
+use std::sync::Arc;
+
 pub fn hello_world() {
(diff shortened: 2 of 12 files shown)


crates/credentials/src/vc.rs
```diff
@@ -107,8 +107,8 @@ impl VerifiableCredential {
 
     pub fn sign(
         &self,
-        bearer_did: &BearerDid,
-        key_selector: &KeySelector,
+        signer: Arc<dyn Signer>,
+        jws_header: JwsHeader,
```
KendallWeihe, line 111: I don't love this byproduct of this design, wherein we have to pass the `jws_header` to the `VerifiableCredential`'s `sign()` method, and the same is true for the `Jwt`'s `sign()` method. Previously, we were able to construct the default `JwsHeader` from the provided `bearer_did`+`key_selector` but since we don't have those anymore, we have to push that concept up the stack. For this reason, you'll see a new method I created in the `jws` crate, `JwsHeader::from_did_document()` which constructs the default `JwsHeader` for the given DID Doc + key selector.

KendallWeihe: Please weigh in if you have alternative ideas. We're prioritizing modularity here over convenience, which is one of the guiding principals, but oof this one kind of hurts.

KendallWeihe: This has triggered a requirement I hadn't previously considered, I don't think our crate design should require the existence of peer dependencies. Moving this PR to a draft stage while I think this through.

KendallWeihe: Created #215 and leaving this code snippet here for now, because this'll probably be what we end up doing here

```rust
pub use keys::{Signer, KeyManagerError};

pub trait CredentialSigner {
    fn sign(&self, payload: &[u8]) -> Result<Vec<u8>, CredentialError>;
}

impl<T> CredentialSigner for T
where
    T: Signer,
{
    fn sign(&self, payload: &[u8]) -> Result<Vec<u8>, CredentialError> {
        T::sign(self, payload).map_err(|e| CredentialError::from(e))
    }
}
```

The idea being: both re-export the underlying `keys` crate in the case the developer wants to use an implementation of the `Signer` from the `keys` crate, and also create a wrapper trait `CredentialSigner` which is compatible with `keys::Signer` but also enables developers to bring-their-own-signer.

decentralgabe: hmm could you show a sample usage here? I'm not clear on why header parameters need to be passed to a generic credential sign - if we are following v1.1. then the decisions have been made for us...and v2 to a lesser degree

maybe the same is not true for other formats

KendallWeihe:
There's also a slightly strange implication with this design change, in that, the `verify()` methods still assume the JWS Header was signed using a DID. Within the `verify()` methods we resolve the DID Document and then make semantic verifications (#182). So by introducing the change in this PR we're enabling the signing of JWS's using an agnostic key signer (AKA not necessarily associated to the DID) but then when we go to verify we assume the JWS was signed with a DID. 

It seems to me if we want to move forward with the proposed changes here then we should also remove the undering DID assumption in the call to `verify()`. And then that raises the question, should we still enable that feature, and if so, where?

crates/credentials/src/vc.rs
KendallWeihe: For sure, if you see the code here 

```rust
impl JwsHeader {
    pub fn from_did_document(
        document: &Document,
        key_selector: &KeySelector,
    ) -> Result<Self, JwsError> {
        let verification_method = document.get_verification_method(key_selector)?;

        Ok(Self {
            alg: verification_method.public_key_jwk.alg.clone(),
            kid: verification_method.id.clone(),
            typ: "JWT".to_string(),
        })
    }
}
```

We're creating a VC-JWT with those three JOSE Header's set, two of which originate from the DID Document. Looking at [the spec](https://www.w3.org/TR/vc-data-model/#proof-formats) I see the language:

> specific JWS-registered header parameter names

@decentralgabe what're your thoughts on what's needed in the VC-JWT JOSE Header's?

decentralgabe: gotcha, I would imagine we construct a signer that takes those parameters as inputs
since any JWT signer with a DID would have the iss and kid fields set as you described

PR #216, opened 2024-05-16
Convert did:dht services to dns records

diehuxx:
Another chunk of did:dht work is tested and ready for review.
Spec: https://did-dht.com/#services


crates/dids/src/method/dht/document_packet/service.rs
```diff
@@ -0,0 +1,222 @@
...
+            .split('#')
+            .last()
+            .ok_or(DocumentPacketError::MissingFragment(self.id.to_string()))?;
+
+        let parts = format!(
+            "id={};t={};se={}",
+            service_id_fragment, self.r#type, self.service_endpoint
+        );
+        let name = Name::new_unchecked(&format!("_s{}._did", idx)).into_owned();
+        let txt_record = TXT::new().with_string(&parts)?.into_owned();
```
decentralgabe, line 88: noting there may be additional service properties for DWNs, defined in the registry
https://did-dht.com/registry/#additional-properties

crates/dids/src/method/dht/document_packet/mod.rs
jiyoontbd, line 1: why is this file named mod.rs?

frankhinek: its a good question @jiyoontbd and one I had contemplated but not opened an issue for.  Perhaps it ought to be a separate issue to ask how we want to approach in `web5-rs`.

The short answer is that it is due to the way module resolution works in the Rust compiler.  Historically, the Rust compiler looked for modules referenced in `use` statements in `src/module_name/mod.rs`.

At least [according to the Rust lang book](https://doc.rust-lang.org/book/ch07-05-separating-modules-into-different-files.html#alternate-file-paths), this is now considered an "older style" and the modern idiomatic way is to use `src/module_name.rs`.  However, there still many Rust devs and libs you'll find that still use the `mod.rs` convention.

jiyoontbd: hmm, so this could be renamed to `src/method/dht/document_packet.rs` if we followed the newer convention? seems like right now this file only contains a document packet error enum though.

crates/dids/src/method/dht/document_packet/service.rs
```diff
@@ -0,0 +1,222 @@
...
+    for entry in text.split(';') {
+        let k_v: Vec<&str> = entry.split('=').collect();
+        if k_v.len() != 2 {
+            return Err(DocumentPacketError::RDataError(
+                "Could not get values from RData text".to_owned(),
+            ));
+        }
+
+        let k = k_v[0].trim().to_string();
+        let v = k_v[1].trim().to_string();
```
jiyoontbd, line 49: don't we also have to delimit the k_v[1] by `,`? or do we just store the values as string here

example: `v=0;vm=k0,k1,k2;auth=k0;asm=k1;inv=k2;del=k2;svc=s0,s1,s2`

vm=k0,k1,k2 should be split into `{ vm: [k0, k1, k2] } ` or is it `{ vm: "k0,k1,k2" }`?

crates/dids/src/method/dht/document_packet/service.rs
```diff
@@ -0,0 +1,222 @@
...
+        )))?;
+    Ok(val.to_string())
+}
+
+impl Service {
+    pub fn to_resource_record(
+        &self,
+        idx: u32,
+    ) -> Result<ResourceRecord<'static>, DocumentPacketError> {
+        let service_id_fragment = self
```
jiyoontbd, line 77: do we want to throw here if split length is more than 2?

crates/dids/src/method/dht/document_packet/mod.rs
frankhinek: Apologies for cluttering the PR... but the other style would look like:

```
└── src
    ├── method
    │   └── dht.rs
    └── dht
        ├── document_packet.rs
        └── document_packet
            └── service.rs
```

Can chat in another issue or Slack.  I don't have a strong feeling one way or the other and it seems like many projects still use the "older" style.

crates/dids/src/method/dht/document_packet/service.rs
jiyoontbd: image

relevant section in did dht spec abt service https://did-dht.com/#services

crates/dids/src/method/dht/document_packet/mod.rs
diehuxx: @jiyoontbd as of this PR that file only contains the error enum, but in the next PR there will be more logic for converting verification methods, identity keys, and a document_to_dns_packet() function to glue it all together. i split this PR from a larger WIP branch that im working from so there's hints of file-structure-yet-to-come.

as for using mod.rs vs module_name.rs, we've been using the mod.rs approach throughout this repo and personally i like. i don't have strong reasoning either way, but for consistency and convenience sake id prefer to keep that convention.

jiyoontbd: That's fair thanks for the context. Perhaps if we get annoyed at having n mod.rs files later we can open a PR  in the future to move to the new naming convention. Thanks for all the explanation!!

crates/dids/src/method/dht/document_packet/service.rs
diehuxx: I decided instead we should return everything after the first `#` instead of returning everything after the final `#`. The URL spec seems to say that addition `#` in a fragment are invalid, but upon running the url crate and node `url` it looks like additional `#` are included as part of the fragment. I don't want to reinvent the wheel, so I'm gonna use the `url` crate and take its default behavior sufficient.

crates/dids/src/method/dht/document_packet/rdata_encoder.rs
diehuxx, line 1: i am NOT psyched about how this files implementation came out. my gut tells me there's a more idiomatic or cleaner way to do this. I need this PR to be done, so i'm gonna call it good enough for now, and look into writing custom `serde` stuff in a follow up PR.

crates/dids/src/document.rs
```diff
@@ -1,5 +1,7 @@
 use jwk::Jwk;
 use serde::{Deserialize, Serialize};
+pub use ssi_core::one_or_many::OneOrMany;
+pub use ssi_dids::ServiceEndpoint;
```
KendallWeihe, line 4: Looks like we aren't using this so can remove, yeah?

crates/dids/src/methods/dht/document_packet/service.rs
```diff
@@ -0,0 +1,237 @@
...
+
+        let txt = TXT::new().with_string("a=b=c;;;").unwrap();
+        let resource_record = ResourceRecord::new(
+            name,
+            simple_dns::CLASS::IN,
+            DEFAULT_TTL,
+            RData::TXT(txt), // Not ';' separated entries
+        );
+
+        let error = Service::from_resource_record("did:ex:abc", resource_record).expect_err("");
```
KendallWeihe, line 212:
```suggestion
        let error = Service::from_resource_record("did:ex:abc", resource_record).expect_err("Expected error because malformed txt");
```

crates/dids/src/methods/dht/document_packet/service.rs
```diff
@@ -0,0 +1,237 @@
...
+
+        let txt = TXT::new().with_string("id=foo;t=bar").unwrap();
+        let resource_record = ResourceRecord::new(
+            name,
+            simple_dns::CLASS::IN,
+            DEFAULT_TTL,
+            RData::TXT(txt), // Not ';' separated entries
+        );
+
+        let error = Service::from_resource_record("did:ex:abc", resource_record).expect_err("");
```
KendallWeihe, line 231:
```suggestion
        let error = Service::from_resource_record("did:ex:abc", resource_record).expect_err("Expected error because missing se");
```

crates/dids/src/methods/dht/document_packet/mod.rs
```diff
@@ -0,0 +1,19 @@
+use simple_dns::SimpleDnsError;
+
+mod rdata_encoder;
```
KendallWeihe, line 3: I appreciate the thoughtfulness of not making this `pub` 👍

crates/dids/src/document.rs
```diff
@@ -63,9 +65,10 @@ impl KeyIdFragment {
 pub struct Service {
     pub id: String,
     #[serde(rename = "type")]
-    pub r#type: String,
+    pub r#type: OneOrMany<String>,
+    // TODO: Update to allow service endpoint as a map?
     #[serde(rename = "serviceEndpoint")]
-    pub service_endpoint: String,
+    pub service_endpoint: OneOrMany<String>,
```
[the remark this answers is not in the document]
KendallWeihe: Sorry re-reading this and I don't think I was clear on this bit:

> And it wouldn't help us with binding because even though we constrain, we still have to support 1 multi-type in [the VC's issuer](https://github.com/TBD54566975/web5-spec/blob/main/spec/vc.md#verifiable-credential-data-model)

What I intended to convey is, we could resort back to only implementing `serviceEndpoints` as an array of strings per our `web5-spec`, but that wouldn't resolve the UDL concern because ultimately we still have one multi-type property with the VC `issuer`.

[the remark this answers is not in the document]
diehuxx: > In case you weren't aware we do constrain our set of types [in web5-spec](https://github.com/TBD54566975/web5-spec/blob/main/spec/did.md#service-data-model)

I was not aware, and I'm very glad you flagged that! I'm gonna update to use Vec<String> rather than OneOrMany<String>. Any opportunity for simplicity.

> that wouldn't resolve the UDL concern because ultimately we still have one multi-type property with the VC issuer.

Gotcha. I'm still in favor using Vec<String> regardless, but this is good context.

crates/dids/src/methods/dht/document_packet/service.rs
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    type Error = DocumentPacketError;
+}
+
+impl Service {
+    pub fn to_resource_record(
+        &self,
+        idx: u32,
+    ) -> Result<ResourceRecord<'static>, DocumentPacketError> {
```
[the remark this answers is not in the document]
diehuxx: good call. updated to what I _think_ accomplishes that

KendallWeihe (approved):

PR #217, opened 2024-05-20
Implementation updated jwt vc payload processing

nitro-neal:
This change allows us to consume a minimum viable VC JWT as well as a full featured VC JWT:

Minimum Viable VC JWT Example:
```
{
  "vc": {
    "@context": [
      "https://www.w3.org/2018/credentials/v1"
    ],
    "type": [
      "VerifiableCredential"
    ]
  },
  "iss": "did:jwk:eyJhbGciOiJFZERTQSIsImNydiI6IkVkMjU1MTkiLCJrdHkiOiJPS1AiLCJ4IjoiTDNjcngtRUlBMkUtbERyN2EtZ250UDdEbVVSMjJyWEg4eERRcWlUWHozMCJ9",
  "sub": "did:jwk:eyJhbGciOiJFZERTQSIsImNydiI6IkVkMjU1MTkiLCJrdHkiOiJPS1AiLCJ4IjoiTDNjcngtRUlBMkUtbERyN2EtZ250UDdEbVVSMjJyWEg4eERRcWlUWHozMCJ9",
  "exp": 1716239889,
  "nbf": 1716238089,
  "jti": "urn:vc:uuid:1ae664b9-9670-4be0-8d64-481df53dc05a"
}
```

we will accept this and dynaimcally build out the VC to fit the VC Data model standards using the claims. (for example vc.issuer will be iss)

We also accept and process a full featured VC JWT like this:
```
{
  "vc": {
    "@context": [
      "https://www.w3.org/2018/credentials/v1"
    ],
    "id": "urn:vc:uuid:dd05667d-8d82-429e-b32b-c557d20a4791",
    "type": [
      "VerifiableCredential"
    ],
    "issuer": "did:jwk:eyJhbGciOiJFZERTQSIsImNydiI6IkVkMjU1MTkiLCJrdHkiOiJPS1AiLCJ4IjoiVGR5SXdHRThlR2dIM3RvZFA0ak1mV0xSaE1kQkg4cW56aW4xZU5RUlpuMCJ9",
    "issuanceDate": 1716239255,
    "expirationDate": 1716241055,
    "credentialSubject": {
      "id": "did:jwk:eyJhbGciOiJFZERTQSIsImNydiI6IkVkMjU1MTkiLCJrdHkiOiJPS1AiLCJ4IjoiVGR5SXdHRThlR2dIM3RvZFA0ak1mV0xSaE1kQkg4cW56aW4xZU5RUlpuMCJ9"
    }
  },
  "iss": "did:jwk:eyJhbGciOiJFZERTQSIsImNydiI6IkVkMjU1MTkiLCJrdHkiOiJPS1AiLCJ4IjoiVGR5SXdHRThlR2dIM3RvZFA0ak1mV0xSaE1kQkg4cW56aW4xZU5RUlpuMCJ9",
  "sub": "did:jwk:eyJhbGciOiJFZERTQSIsImNydiI6IkVkMjU1MTkiLCJrdHkiOiJPS1AiLCJ4IjoiVGR5SXdHRThlR2dIM3RvZFA0ak1mV0xSaE1kQkg4cW56aW4xZU5RUlpuMCJ9",
  "exp": 1716241055,
  "nbf": 1716239255,
  "jti": "urn:vc:uuid:dd05667d-8d82-429e-b32b-c557d20a4791"
}
```

### DIFF
--- a/crates/credentials/src/verifiable_credential.rs
+++ b/crates/credentials/src/verifiable_credential.rs
@@ -130,22 +130,22 @@ pub struct VerifiableCredential {
     pub credential_subject: CredentialSubject,
 }
 
-impl VerifiableCredential {
-    fn into_jwt_payload_verifiable_credential(self) -> JwtPayloadVerifiableCredential {
+impl From<VerifiableCredential> for JwtPayloadVerifiableCredential {
+    fn from(credential: VerifiableCredential) -> Self {
         JwtPayloadVerifiableCredential {
-            context: self.context,
-            id: Some(self.id),
-            r#type: self.r#type,
-            issuer: Some(self.issuer),
-            issuance_date: Some(self.issuance_date),
-            expiration_date: self.expiration_date,
-            credential_subject: Some(self.credential_subject),
+            context: credential.context,
+            id: Some(credential.id),
+            r#type: credential.r#type,
+            issuer: Some(credential.issuer),
+            issuance_date: Some(credential.issuance_date),
+            expiration_date: credential.expiration_date,
+            credential_subject: Some(credential.credential_subject),
         }
     }
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
-struct JwtPayloadVerifiableCredential {
+pub struct JwtPayloadVerifiableCredential {
     #[serde(rename = "@context")]
     context: Vec<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
@@ -162,16 +162,16 @@ struct JwtPayloadVerifiableCredential {
     credential_subject: Option<CredentialSubject>,
 }
 
-impl JwtPayloadVerifiableCredential {
-    fn into_verifiable_credential(self) -> VerifiableCredential {
+impl From<JwtPayloadVerifiableCredential> for VerifiableCredential {
+    fn from(payload: JwtPayloadVerifiableCredential) -> Self {
         VerifiableCredential {
-            context: self.context,
-            id: self.id.unwrap_or_default(),
-            r#type: self.r#type,
-            issuer: self.issuer.unwrap(),
-            issuance_date: self.issuance_date.unwrap_or_default(),
-            expiration_date: self.expiration_date,
-            credential_subject: self.credential_subject.unwrap(),
+            context: payload.context,
+            id: payload.id.unwrap_or_default(),
+            r#type: payload.r#type,
+            issuer: payload.issuer.unwrap(),
+            issuance_date: payload.issuance_date.unwrap_or_default(),
+            expiration_date: payload.expiration_date,
+            credential_subject: payload.credential_subject.unwrap(),
         }
     }
 }
@@ -226,7 +226,7 @@ impl VerifiableCredential {
                 expiration: self.expiration_date,
                 ..Default::default()
             },
-            vc_payload: self.clone().into_jwt_payload_verifiable_credential(),
+            vc_payload: self.clone().into(),
         };
 
         let jwt = Jwt::sign(bearer_did, key_selector, None, &claims)?;
@@ -277,32 +277,43 @@ impl VerifiableCredential {
             }
         }
 
-        if let Some(expiration_date) = &vc_payload.expiration_date {
-            if exp.is_none() {
-                return Err(CredentialError::MisconfiguredExpirationDate(
-                    "VC has expiration date but no exp in registered claims".to_string(),
-                ));
-            }
+        match vc_payload.expiration_date {
+            Some(ref vc_payload_expiration_date) => match exp {
+                None => {
+                    return Err(CredentialError::MisconfiguredExpirationDate(
+                        "VC has expiration date but no exp in registered claims".to_string(),
+                    ));
+                }
+                Some(exp) => {
+                    if vc_payload_expiration_date != &exp {
+                        return Err(CredentialError::ExpirationDateMismatch);
+                    }
 
-            if expiration_date != &exp.unwrap() {
-                return Err(CredentialError::IssuanceDateMismatch);
+                    let now = Utc::now().timestamp();
+                    if now > exp {
+                        return Err(CredentialError::CredentialExpired(
+                            "The verifiable credential has expired".to_string(),
+                        ));
+                    }
+                }
+            },
+            None => {
+                if let Some(exp) = exp {
+                    let now = Utc::now().timestamp();
+                    if now > exp {
+                        return Err(CredentialError::CredentialExpired(
+                            "The verifiable credential has expired".to_string(),
+                        ));
+                    }
+                }
             }
         }
 
-        if exp.is_some() {
-            let now = Utc::now().timestamp();
-            if now > exp.unwrap() {
-                return Err(CredentialError::CredentialExpired(
-                    "The verifiable credential has expired".to_string(),
-                ));
-            }
-        }
-
-        let mut vc_issuer = Issuer::String(iss.clone());
-
-        if let Some(Issuer::Object(_)) = vc_payload.issuer.as_ref() {
-            vc_issuer = vc_payload.issuer.clone().unwrap();
-        }
+        let vc_issuer = match vc_payload.issuer {
+            Some(Issuer::Object(ref issuer_obj)) => Issuer::Object(issuer_obj.clone()),
+            Some(Issuer::String(ref issuer_str)) => Issuer::String(issuer_str.clone()),
+            None => Issuer::String(iss),
+        };
 
         let vc_credential_subject =
             vc_payload
@@ -314,16 +325,16 @@ impl VerifiableCredential {
                 });
 
         let vc = VerifiableCredential {
-            context: vc_payload.context.clone(),
-            id: jti.clone(),
-            r#type: vc_payload.r#type.clone(),
-            issuer: vc_issuer.clone(),
+            context: vc_payload.context,
+            id: jti,
+            r#type: vc_payload.r#type,
+            issuer: vc_issuer,
             issuance_date: nbf,
             expiration_date: exp,
             credential_subject: vc_credential_subject,
         };
 
-        validate_vc_data_model(&vc).map_err(CredentialError::ValidationError)?;
+        validate_vc_data_model(&vc)?;
 
         Ok(vc)
     }
@@ -331,7 +342,7 @@ impl VerifiableCredential {
     pub fn decode(jwt: &str) -> Result<Self, CredentialError> {
         let jwt_decoded = Jwt::decode::<VcJwtClaims>(jwt)?;
 
-        let vc = jwt_decoded.claims.vc_payload.into_verifiable_credential();
+        let vc = jwt_decoded.claims.vc_payload.into();
         Ok(vc)
     }
 }
--- a/crates/credentials/src/verifiable_credential.rs
+++ b/crates/credentials/src/verifiable_credential.rs
@@ -17,71 +17,20 @@ const BASE_TYPE: &str = "VerifiableCredential";
 
 #[derive(thiserror::Error, Debug)]
 pub enum CredentialError {
-    // JWT specific errors
     #[error(transparent)]
     JwtError(#[from] JwtError),
     #[error(transparent)]
     JwsError(#[from] JwsError),
-
-    // Credential-specific validation errors
-    #[error("Expiration date mismatch")]
-    ExpirationDateMismatch,
-    #[error("Credential expired: {0}")]
-    CredentialExpired(String),
-    #[error("Issuer mismatch")]
-    IssuerMismatch,
-    #[error("Issuance date is in the future")]
-    IssuanceDateInFuture,
-    #[error("Issuance date mismatch")]
-    IssuanceDateMismatch,
-    #[error("Credential ID mismatch")]
-    IdMismatch,
-    #[error("Subject mismatch")]
-    SubjectMismatch,
-    #[error("Missing issuer")]
-    MissingIssuer,
-    #[error("Missing issuance date")]
-    MissingIssuanceDate,
-    #[error("Missing subject")]
-    MissingSubject,
-    #[error("Missing credential ID")]
-    MissingId,
-    #[error("Misconfigured expiration date: {0}")]
+    #[error("missing claim: {0}")]
+    MissingClaim(String),
+    #[error("claim mismatch: {0}")]
+    ClaimMismatch(String),
+    #[error("misconfigured expiration date: {0}")]
     MisconfiguredExpirationDate(String),
-
-    // Data model validation error
-    #[error("Missing Iss")]
-    MissingIss,
-    #[error("Missing Nbf")]
-    MissingNbf,
-    #[error("Missing Sub")]
-    MissingSub,
-    #[error("Missing Jti")]
-    MissingJti,
-    #[error("Validation error: {0}")]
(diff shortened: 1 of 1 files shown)


crates/credentials/src/verifiable_credential.rs
```diff
@@ -130,6 +130,52 @@ pub struct VerifiableCredential {
     pub credential_subject: CredentialSubject,
 }
 
+impl VerifiableCredential {
```
diehuxx, line 133: What do you think of using the `From` trait here? It automatically implements the corresponding `into()` method.

```suggestion
impl From<VerifiableCredential> for JwtPayloadVerifiableCredential {
```

crates/credentials/src/verifiable_credential.rs
```diff
@@ -130,6 +130,52 @@ pub struct VerifiableCredential {
...
+    issuer: Option<Issuer>,
+    #[serde(skip_serializing_if = "Option::is_none", rename = "issuanceDate")]
+    issuance_date: Option<i64>,
+    #[serde(skip_serializing_if = "Option::is_none", rename = "expirationDate")]
+    expiration_date: Option<i64>,
+    #[serde(skip_serializing_if = "Option::is_none", rename = "credentialSubject")]
+    credential_subject: Option<CredentialSubject>,
+}
+
+impl JwtPayloadVerifiableCredential {
```
diehuxx, line 165: ditto https://github.com/TBD54566975/web5-rs/pull/217/files#r1608933430, we could use the `From` trait

```suggestion
impl From<JwtPayloadVerifiableCredential> for VerifiableCredential {
```

crates/credentials/src/verifiable_credential.rs
```diff
@@ -206,64 +252,87 @@ impl VerifiableCredential {
...
+                return Err(CredentialError::IssuanceDateMismatch);
+            }
         }
 
-        // if exp exists, make sure there is not a mismatch and assign it to vc expiration date
-        // if vc expiration date exists and exp does not, throw a misconfigured exp error
         if exp.is_some() {
-            if vc.expiration_date.is_some() {
-                if vc.expiration_date != exp {
-                    return Err(CredentialError::ExpirationDateMismatch);
```
diehuxx, line 241: With this line being deleted, we don't use `CredentialError::ExpirationDateMismatch` anymore. Should we remove that variant? Or are we now missing an if-statement in here that should use it

crates/credentials/src/verifiable_credential.rs
```diff
@@ -130,6 +130,52 @@ pub struct VerifiableCredential {
...
+    credential_subject: Option<CredentialSubject>,
+}
+
+impl JwtPayloadVerifiableCredential {
+    fn into_verifiable_credential(self) -> VerifiableCredential {
+        VerifiableCredential {
+            context: self.context,
+            id: self.id.unwrap_or_default(),
+            r#type: self.r#type,
+            issuer: self.issuer.unwrap(),
```
diehuxx, line 171: `unwrap()` is scary. that's a , i.e. the program crashes. we should probably return a `Result` instead

crates/credentials/src/verifiable_credential.rs
```diff
@@ -206,64 +252,87 @@ impl VerifiableCredential {
...
-        } else if vc.issuance_date != nbf {
-            return Err(CredentialError::IssuanceDateMismatch);
+        if let Some(expiration_date) = &vc_payload.expiration_date {
+            if exp.is_none() {
+                return Err(CredentialError::MisconfiguredExpirationDate(
+                    "VC has expiration date but no exp in registered claims".to_string(),
+                ));
+            }
+
+            if expiration_date != &exp.unwrap() {
```
diehuxx, line 287: ditto https://github.com/TBD54566975/web5-rs/pull/217/files#r1609142601

crates/credentials/src/verifiable_credential.rs
```diff
@@ -206,64 +252,87 @@ impl VerifiableCredential {
...
-                let now = Utc::now().timestamp();
-                if now > exp.unwrap() {
-                    return Err(CredentialError::CredentialExpired(
-                        "The verifiable credential has expired".to_string(),
-                    ));
-                }
-
-                vc.expiration_date = exp;
+            let now = Utc::now().timestamp();
+            if now > exp.unwrap() {
```
diehuxx, line 294: ditto https://github.com/TBD54566975/web5-rs/pull/217/files#r1609142601

crates/credentials/src/verifiable_credential.rs
```diff
@@ -206,64 +252,87 @@ impl VerifiableCredential {
...
-        } else if vc.expiration_date.is_some() {
-            return Err(CredentialError::MisconfiguredExpirationDate(
-                "VC has expiration date but no exp in registered claims".to_string(),
-            ));
         }
 
+        let mut vc_issuer = Issuer::String(iss.clone());
+
+        if let Some(Issuer::Object(_)) = vc_payload.issuer.as_ref() {
+            vc_issuer = vc_payload.issuer.clone().unwrap();
```
diehuxx, line 304: ditto https://github.com/TBD54566975/web5-rs/pull/217/files#r1609142601

crates/credentials/src/verifiable_credential.rs
diehuxx: ah i see you're checking `if exp.is_none()`. i'd prefer if we use a `match` statement, because seeing an `unwrap()` outside of tests sets off mental alarm bells.

crates/credentials/src/verifiable_credential.rs
diehuxx: in this since we do `if exp.is_some()`, I'd prefer if we did `if let Some(exp) = exp`

crates/credentials/src/verifiable_credential.rs
```diff
@@ -206,64 +252,87 @@ impl VerifiableCredential {
...
+            context: vc_payload.context.clone(),
+            id: jti.clone(),
+            r#type: vc_payload.r#type.clone(),
+            issuer: vc_issuer.clone(),
+            issuance_date: nbf,
+            expiration_date: exp,
+            credential_subject: vc_credential_subject,
+        };
+
         validate_vc_data_model(&vc).map_err(CredentialError::ValidationError)?;
```
diehuxx, line 326:
```suggestion
        validate_vc_data_model(&vc)?;
```

crates/credentials/src/verifiable_credential.rs
```diff
@@ -206,64 +252,87 @@ impl VerifiableCredential {
...
+                });
+
+        let vc = VerifiableCredential {
+            context: vc_payload.context.clone(),
+            id: jti.clone(),
+            r#type: vc_payload.r#type.clone(),
+            issuer: vc_issuer.clone(),
+            issuance_date: nbf,
+            expiration_date: exp,
+            credential_subject: vc_credential_subject,
```
diehuxx, line 323: i think all of these `clone()`s can be omitted. those values are getting dropped at the end of the function anyway

crates/credentials/src/verifiable_credential.rs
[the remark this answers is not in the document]
diehuxx: if you feel like really finessing it, you could combine this assignment with the `if let Some(issuer) = &vc_payload.issuer` that happens much further up in the function.

diehuxx:
Suggestion for the `CredentialError`, that I probably should have brought up in an earlier PR:

Both `CredentialError` and `CredentialDataModelValidationError` have a lot of variants, and some of them aren't even used, which tells me that we should pare them both down a lot.

A lot of variants in `CredentialError` amount to "this field is missing" or "these fields mismatch". What do you think of consolidating each of those concepts into one error variant each? If you prefer, we can have `MissingVcField` and `MissingJwtField` be separate, but I think it would also be okay to just have one `MissingField`.

```rust
#[derive(thiserror::Error, Debug)]
pub enum CredentialError {
    // JWT specific errors
    ...

    // Credential-specific validation errors
    #[error("Credential expired: {0}")]
    CredentialExpired(String),
    #[error("Issuance date is in the future")]
    IssuanceDateInFuture,
    #[error("Misconfigured expiration date: {0}")]
    MisconfiguredExpirationDate(String),
    #[error("Mismatch between VC field {0} and JWT field {1}")]
    Mismatch(String, String),
    #[error("Missing VC field {0}")]
    MissingVcField(String),
    #[error("Missing JWT field {0}")]
    MissingJwtField(String).
}
```

I would also completely remove `CredentialDataModelValidationError`. Most of its variants have some equivalent in `CredentialError` and we only propagate `CredentialDataModelValidationError` wrapped inside of `CredentialError::ValidationError`.

crates/credentials/src/verifiable_credential.rs
diehuxx, line 326: I wonder if `validate_data_model` would be useful as a public method on `VerifiableCredential`

crates/credentials/src/verifiable_credential.rs
```diff
@@ -130,6 +130,52 @@ pub struct VerifiableCredential {
...
+            issuer: Some(self.issuer),
+            issuance_date: Some(self.issuance_date),
+            expiration_date: self.expiration_date,
+            credential_subject: Some(self.credential_subject),
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+struct JwtPayloadVerifiableCredential {
```
diehuxx, line 148: Should this be a `pub struct`? It's used as a field of `pub struct VcJwtClaims`. It would be weird for a consumer to work with a `VcJwtClaims` but not be able to work easily with one of its fields. If it should be `pub`, can you add a doc comment explaining why `JwtPayloadVerifiableCredential` is a distinct type from `VerifiableCredential` and how the two should be used distinctly.

crates/credentials/src/verifiable_credential.rs
nitro-neal: oh thats pretty cool, yea I'm about this (less code, but less flexibility?) Seems like we dont need any special case stuff so I'll add it

crates/credentials/src/verifiable_credential.rs
nitro-neal: Yup that's how we had it many moons ago, but decided it was too hard for the end user to do vc.verify() vs vc.validate()

so was going with that latest api "spec"

until we have a more concrete api spec that @KendallWeihe has been working on and / or we just really want to do it lol.

crates/credentials/src/verifiable_credential.rs
nitro-neal: updated!

crates/credentials/src/verifiable_credential.rs
nitro-neal: updated!

crates/credentials/src/verifiable_credential.rs
nitro-neal: updated to get rid of all unwraps 👍

crates/credentials/src/verifiable_credential.rs
nitro-neal: yup probably so! I'll make it pub

crates/credentials/src/verifiable_credential.rs
nitro-neal: no more unwraps 👍

crates/credentials/src/verifiable_credential.rs
diehuxx: the linter won't catch unused enum variants if the enum is public. The reason is the linter thinks "hmm we're not using that variant in this file, but maybe a consumer of this module/crate is using the variant somewhere?"

crates/credentials/src/verifiable_credential.rs
```diff
@@ -17,71 +17,20 @@ const BASE_TYPE: &str = "VerifiableCredential";
 
 #[derive(thiserror::Error, Debug)]
 pub enum CredentialError {
-    // JWT specific errors
     #[error(transparent)]
     JwtError(#[from] JwtError),
     #[error(transparent)]
     JwsError(#[from] JwsError),
```
diehuxx, line 23: doesn't look like we're using this anywhere

crates/credentials/src/verifiable_credential.rs
```diff
@@ -130,6 +79,52 @@ pub struct VerifiableCredential {
...
+impl From<JwtPayloadVerifiableCredential> for VerifiableCredential {
+    fn from(payload: JwtPayloadVerifiableCredential) -> Self {
+        VerifiableCredential {
+            context: payload.context,
+            id: payload.id.unwrap_or_default(),
+            r#type: payload.r#type,
+            issuer: payload.issuer.unwrap(),
+            issuance_date: payload.issuance_date.unwrap_or_default(),
+            expiration_date: payload.expiration_date,
+            credential_subject: payload.credential_subject.unwrap(),
```
diehuxx, line 123: a couple more `.unwrap()`s to kill
```suggestion
            issuer: payload.issuer.unwrap_or_default(),
            issuance_date: payload.issuance_date.unwrap_or_default(),
            expiration_date: payload.expiration_date,
            credential_subject: payload.credential_subject.unwrap_or_default(),
```

crates/credentials/src/verifiable_credential.rs
```diff
@@ -180,7 +175,7 @@ impl VerifiableCredential {
                 expiration: self.expiration_date,
                 ..Default::default()
             },
-            vc: self.clone(),
+            vc_payload: self.clone().into(),
```
diehuxx, line 178: nice! love getting to use '.into()`

crates/credentials/src/verifiable_credential.rs
```diff
@@ -190,121 +185,168 @@ impl VerifiableCredential {
...
+        let vc_issuer = match vc_payload.issuer {
+            Some(Issuer::Object(ref issuer_obj)) => Issuer::Object(issuer_obj.clone()),
+            Some(Issuer::String(ref issuer_str)) => Issuer::String(issuer_str.clone()),
+            None => Issuer::String(iss),
+        };
+
+        let vc_credential_subject =
+            vc_payload
+                .credential_subject
+                .clone()
```
diehuxx, line 269: we can remove this `.clone()` too

crates/credentials/src/verifiable_credential.rs
diehuxx: if we don't want to use `default()` for those fields, we could return `Err` instead. In that case we would use the `TryFrom` trait instead of the `From` trait.

crates/credentials/src/verifiable_credential.rs
```diff
@@ -130,6 +79,52 @@ pub struct VerifiableCredential {
     pub credential_subject: CredentialSubject,
 }
 
+impl From<VerifiableCredential> for JwtPayloadVerifiableCredential {
+    fn from(credential: VerifiableCredential) -> Self {
+        JwtPayloadVerifiableCredential {
+            context: credential.context,
+            id: Some(credential.id),
```
diehuxx, line 86: I'm not sure if it matters, but should this be `None` if `id.is_empty()`? Same question for `issuer`, `issuance_date`, and `credential_subject`.

diehuxx (commented):
Really close! Just a couple more things. Great work so far!

diehuxx (approved):
Noice! Last thing left is to remove `JwsError` from `CredentialError`. Other than that, this is looking good

Issue #205, opened 2024-05-14
Add singular cryptography interfaces for generating, signing and verifying


KendallWeihe:
In a ton of places throughout the codebase, wherever we make a call to on of the `crypto` crates' `generate()`/`sign()`/`verify()` functions we always have to have a `match` statement. For example, 

```rust
        let private_key = Arc::new(match curve {
            Curve::Ed25519 => Ed25519::generate(),
            Curve::Secp256k1 => Secp256k1::generate(),
        }?);
```

This is an obvious place where DRY and the convenience would save us a lot of lines of code (and overhead). We should create singular interfaces for this set of functionality, within the `crypto` crate, perhaps multiple different functions depending on what the required set of parameters are (rust doesn't support function overloading, so prefer to use required parameters with named functions; for example `sign_crv(curve: Curve)` and `sign_crv_str(curve: &str)`).

KendallWeihe:
oh interesting, I can see @diehuxx you're already thinking about this with #156, nice!

PR #218, opened 2024-05-22
#205 Convert Curve to/from string

diehuxx:
addresses #205

### DIFF
--- a/crates/crypto/src/lib.rs
+++ b/crates/crypto/src/lib.rs
@@ -41,8 +41,8 @@ impl FromStr for Curve {
     type Err = CryptoError;
 
     fn from_str(input: &str) -> Result<Self, CryptoError> {
-        match input {
-            "Ed25519" => Ok(Curve::Ed25519),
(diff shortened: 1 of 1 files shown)


crates/crypto/src/lib.rs
```diff
@@ -22,15 +27,63 @@ pub enum CryptoError {
...
     Ed25519,
 }
 
+impl FromStr for Curve {
+    type Err = CryptoError;
+
+    fn from_str(input: &str) -> Result<Self, CryptoError> {
+        match input {
+            "Ed25519" => Ok(Curve::Ed25519),
+            "secp256k1" => Ok(Curve::Secp256k1),
```
KendallWeihe, line 46: I wonder if we should be case-agnostic 🤔

KendallWeihe (approved):
nice! love this!

KendallWeihe:
let's leave #205 open tho because we may have to go back through and update all the relevant areas where we can leverage these new handy dandy converters, yeah?

PR #219, opened 2024-05-24
Convert did:dht verification method to DNS record

diehuxx:
Another PR on the way to did:dht. This PR implements [the section of the spec](https://did-dht.com/#verification-methods) related to converting verification methods to DNS records.

Implementation Note: Since there are upcoming changes to the API surface of the crypto and keys package, I made the most minimal changes to crypto and keys I could to unblock converting VMs to DNS records. After we decide how to refactor crypto and keys, we can consider updating `dht/document_packet/verification_method.rs` accordingly.

In follow up PR, `to_resource_record()` and `from_resource_record()` will be utilized for converting an entire DID document to and from a DNS packet.

### DIFF
--- a/crates/crypto/src/secp256k1.rs
+++ b/crates/crypto/src/secp256k1.rs
@@ -11,6 +11,37 @@ use k256::{
 
 pub struct Secp256k1;
 
+impl Secp256k1 {
+    pub fn extract_public_key(jwk: &Jwk) -> Result<Vec<u8>, CryptoError> {
+        let decoded_x = general_purpose::URL_SAFE_NO_PAD.decode(&jwk.x)?;
+        let decoded_y = general_purpose::URL_SAFE_NO_PAD.decode(
+            jwk.y
+                .as_ref()
+                .ok_or(CryptoError::PublicKeyFailure("missing y".to_string()))?,
+        )?;
+
+        let mut pk_bytes = Vec::with_capacity(1 + decoded_x.len() + decoded_y.len());
+        pk_bytes.push(0x04); // Uncompressed point indicator
+        pk_bytes.extend_from_slice(&decoded_x);
+        pk_bytes.extend_from_slice(&decoded_y);
+
+        Ok(pk_bytes)
+    }
+
+    pub fn from_public_key(public_key: &[u8]) -> Result<Jwk, CryptoError> {
+        let x_bytes = &public_key[1..33];
+        let y_bytes = &public_key[33..65];
+        Ok(Jwk {
+            alg: "ES256K".to_string(),
+            kty: "EC".to_string(),
+            crv: "secp256k1".to_string(),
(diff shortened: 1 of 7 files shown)


crates/dids/src/methods/dht/document_packet/verification_method.rs
```diff
@@ -0,0 +1,295 @@
...
+
+    type Error = DocumentPacketError;
+}
+
+impl VerificationMethod {
+    pub fn to_resource_record(
+        &self,
+        did_uri: &str,
+        idx: u32,
+    ) -> Result<ResourceRecord<'static>, DocumentPacketError> {
```
nitro-neal, line 43: why does it have to be static?

crates/crypto/src/secp256k1.rs
```diff
@@ -11,6 +11,37 @@ use k256::{
...
+    pub fn extract_public_key(jwk: &Jwk) -> Result<Vec<u8>, CryptoError> {
+        let decoded_x = general_purpose::URL_SAFE_NO_PAD.decode(&jwk.x)?;
+        let decoded_y = general_purpose::URL_SAFE_NO_PAD.decode(
+            jwk.y
+                .as_ref()
+                .ok_or(CryptoError::PublicKeyFailure("missing y".to_string()))?,
+        )?;
+
+        let mut pk_bytes = Vec::with_capacity(1 + decoded_x.len() + decoded_y.len());
+        pk_bytes.push(0x04); // Uncompressed point indicator
```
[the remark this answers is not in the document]
diehuxx: For sure I'll add a few more words there. here's some more context for the cryptographically curious:

sep256k1 public keys are composed of an `x` and `y` coordinate on an elliptic curve. Since elliptic curves are symmetrical, you can omit the `y` coordinate, and just have a single bit that notes if y is positive or negative. This is a "compressed" public key which is 33 bytes long. An "uncompressed" public key has the full y coordinate, making the public key 65 bytes long. To denote that the public key is uncompressed, it's standard to prefix it with the byte `0x04`.

When you're extremely space contrained, such as on bitcoin, shaving off 32 bytes can make a real difference.

This [blog post](https://davidederosa.com/basic-blockchain-programming/elliptic-curve-keys/#:~:text=Public%20keys%20are%2064%20bytes,is%20based%20on%20modular%20arithmetic.) has a pretty friendly explanation.

For an even deeper dive, I'd recommend [Chapter 3 of Jimmy Song's Programming Bitcoin](https://github.com/jimmysong/programmingbitcoin/blob/master/ch03.asciidoc).

crates/dids/src/methods/dht/document_packet/verification_method.rs
diehuxx: ah, that was a mistake. updated to remove the lifetime annotation.

PR #225, opened 2024-06-03
One crate

diehuxx:
Consolidate all crates into one. In a future PR, we will probably want to more neatly export specific items and un-nest some items from their deeply nested modules, rather than just `pub mod`ing everything. But this PR is just to move us to one crate with minimal changes.


crates/web5/src/keys/key_manager/local_key_manager.rs
```diff
@@ -106,7 +106,7 @@ impl Default for LocalKeyManager {
 
 #[cfg(test)]
 mod tests {
-    use crate::key::Key;
+    use crate::keys::key::Key;
```
nitro-neal, line 109: later on we can think of a different names if people see this when they import from kotlin or something 😆

crates/web5/Cargo.toml
```diff
@@ -8,23 +8,31 @@ license-file.workspace = true
...
+k256 = { version = "0.13.3", features = ["ecdsa", "jwk"] }
 rand = { workspace = true }
+regex = "1.10.4"
 serde = { workspace = true }
 serde_json = { workspace = true }
 serde_with = { workspace = true }
+sha2 = "0.10.8"
+simple-dns = "0.7.0"
+ssi-core = "0.1.0"
+ssi-dids = "0.1.1"
```
nitro-neal, line 29: we still use this?

crates/web5/src/keys/key_manager/local_key_manager.rs
diehuxx: [💯](https://emojipedia.org/hundred-points)

nitro-neal (approved):
Awesome, wasn't as bad as I thought it would be.

I wonder how large our binding is for the entire web5 create (and js wasm size)

May be worth to do some work and record the binding size when we get back into bindings. can see the starting baseline size

crates/web5/Cargo.toml
diehuxx: yep, we're using ssi-dids and ssi-core for our did:web and did:jwk.

bindings/uniffi/Cargo.toml
```diff
@@ -7,20 +7,14 @@ repository.workspace = true
...
-keys = { path = "../../crates/keys" }
+web5 = { path = "../../crates/web5" }
 uniffi = { version = "0.27.1", features = ["cli"] }
 
 [build-dependencies]
 uniffi = { version = "0.27.1", features = ["build"] }
 
 [lib]
 crate-type = ["cdylib"]
-name = "web5"
```
KendallWeihe, line 23: FYI this is relevant to the binding scripts https://github.com/TBD54566975/web5-rs/pull/227

crates/web5/src/credentials/presentation_definition.rs
```diff
@@ -237,7 +237,7 @@ impl JsonSchemaBuilder {
 mod tests {
     use std::collections::HashSet;
 
-    use test_helpers::TestVectorFile;
+    use crate::test_helpers::TestVectorFile;
```
KendallWeihe, line 240: @diehuxx thoughts on

- `crates/`
    - `test_helpers`
    - `web5`

?

KendallWeihe: Or, thoughts on simply 

- `core/`
    - `src/`
    - `Cargo.toml`
    - etc.

Meaning, just lean hard into only a single crate? 

Any of the above three are fine because the compiler will handle performance matters, so it's purely cosmetic, just ideating

PR #229, opened 2024-06-07
RFC-0001 Standard Web5 API Design (APID) Document v0.1.0

KendallWeihe:
This is a first version of the [Deliverable 1: Web5 API Design (APID)](https://github.com/TBD54566975/technical-steering-committee/pull/11) as specified in the Rust Core WG Charter. For context, I am posting this PR for review prior-to merging the linked TSC Charter Proposal, because of soft consensus already established and with the intent of giving reviewers ample time to review both; as stated in the linked PR:

> I share here, with respect to the charter, because the concept of a standard API Design is instrumental (and novel) to the proposed charter, and so some semblance of familiarity may be beneficial in review of this charter.

In other words, since the APID is instrumental to the success of the Rust Core WG, reviewing both in combination offers the fairest degree of judgement.

This is incomplete. The reason for merging is to unblock development on the pieces which are completed. Incomplete items are defined in the _Design Details_ section under _APID v0.1.0_. Copy and pasting here for convenience:

> `v0.1.0` of the APID excludes the following features (in completeness), all of which should be added in subsequent versions:
> 
> - Doc comments.
> - Test vectors.
> - Examples.
> - Generalized object type (AKA a JSON Object).
> - Errors.
> - JSON Serialized key-value key name casing (we have both snake case and camel case currently throughout the data members, because the Custom DSL doesn't offer a feature for overriding the key-value's key casing).
> - Namespacing.
> 
> `v0.1.0` of the APID defines an incomplete design for the following features:
> 
> - `VerifiableCredential`: The `credentialSubject` data member on the `VerifiableCredential` is intended to be of type `Object` but the Custom DSL does not currently support such a primitive type. The requirement of the type as an `Object` originates [from the `web5-spec`](https://github.com/TBD54566975/web5-spec/blob/main/spec/vc.md#verifiable-credential-data-model). A subsequent version of the Custom DSL and the APID is expected to define this feature.
> - `InMemoryKeyManager`: The `InMemoryKeyManager` class offers sufficient support for enablement of downstream DIDs and VC features, however, there is room for additional methods on the key manager, as well as a secondary key manager altogether to showcase the degree of composability enabled in the surround design.

---

Reviewers, please consider the following proposal. 

## Rename `Did` to `Identifier`

In the APID I have a class named `Did`, and this class represents the matters from [Section 3. Identifier](https://www.w3.org/TR/did-core/#identifier) of the DID Core spec. Historically, in many implementations, we refer to this concept as `Did`, and so that is what I have defined in the APID. However, considering the name of the section in the specification, and as to not conflate a "DID" (which is an amorphous and broad concept which includes more than merely the identifier) and an "Identifier", I propose we name this `Identifier`. (We actually currently do this in [the existing `web5-rs` implementation.](https://github.com/TBD54566975/web5-rs/blob/3551742a15dd516ceee8a5aa58418e82a7377133/crates/web5/src/dids/identifier.rs#L16))


docs/API_DESIGN.md
```diff
@@ -0,0 +1,652 @@
...
+  METHOD verify(message: []byte, signature: []byte): bool
+```
+
+## `Ed25519Generator`
+
+```pseudocode!
+/// Generates private key material for Ed25519.
+CLASS Ed25519Generator
+  /// Generate the private key material; return Jwk includes private key material.
+  STATIC METHOD generate(): Jwk
```
KendallWeihe, line 141:
```suggestion
  STATIC METHOD generate_key_material(): Jwk
```

?

docs/API_DESIGN.md
```diff
@@ -0,0 +1,652 @@
...
+  METHOD generate_key_material(): Jwk
+
+  /// Returns the Ed25519Signer for the given public key.
+  METHOD get_signer(public_key: Jwk): Ed25519Signer
+
+  /// For importing keys which may be stored somewhere such as environment variables. Return Jwk is the public key for the given private key.
+  METHOD import_key(private_key: Jwk): Jwk
+```
+
+# Digital Signature Algorithm's (DSA)
```
decentralgabe, line 105: I'd suggest a more generic name like `Cryptography` or even just `Crypto` since the code here is already (and will continue to be) broader than _signature algorithms_. Key generation now - and in the future things like encryption or key agreement will make this naming obsolete.

docs/API_DESIGN.md
```diff
@@ -0,0 +1,652 @@
...
+
+## `Dsa`
+
+```pseudocode!
+/// The set of Digital Signature Algorithm's natively supported within this SDK.
```
decentralgabe, line 110:
```suggestion
/// The set of cryptographic primitives natively supported within this SDK.
```

docs/API_DESIGN.md
```diff
@@ -0,0 +1,652 @@
...
+
+## `NamedIssuer`
+
+```pseudocode!
+CLASS NamedIssuer
+  PUBLIC DATA id: string
+  PUBLIC DATA name: string
+```
+
+## `VerifiableCredential`
```
decentralgabe, line 539: it may make sense to namespace this, since we'll soon need to account for differnet types of 'verifiable credentials' or the more generic 'digital credentials' - for example: SD-JWT, SD-JWT-VC, mDL/mDoc, W3C VCDM v1 (what we currently support), W3C VCDM v2

decentralgabe (commented):
AWESOMe work!

docs/API_DESIGN.md
KendallWeihe: The _Custom DSL_ doesn't currently offer prescriptions with respect to namespacing, but I think it may be prudent to consider. I've added namespacing as a missing concept, and I've moved the above section as a sub-section under `Cryptography`, and I've done the same for the VC Data Model 1.1 in your comment below. Thanks!

docs/API_DESIGN.md
KendallWeihe: Moved under `Data Model 1.1` markdown section

docs/API_DESIGN.md
decentralgabe: related to [this comment](https://github.com/TBD54566975/web5-rs/pull/229#discussion_r1638155385) -- this API should follow that logic. it's sensible to have an API for DSA, though it should just be for DSA. Key generation should be separate. Same with encryption in the future - and other cryptographic functions.

decentralgabe: which I've noticed you've already done 😄

docs/API_DESIGN.md
```diff
@@ -0,0 +1,658 @@
...
+  /// Returns the Ed25519Signer for the given public key.
+  METHOD get_signer(public_key: Jwk): Ed25519Signer
+
+  /// For importing keys which may be stored somewhere such as environment variables. Return Jwk is the public key for the given private key.
+  METHOD import_key(private_key: Jwk): Jwk
+```
+
+# Cryptography
+
+## Digital Signature Algorithm's (DSA)
```
angiejones, line 109:
```suggestion
## Digital Signature Algorithms (DSA)
```

docs/API_DESIGN.md
```diff
@@ -0,0 +1,658 @@
...
+
+### `Dsa`
+
+```pseudocode!
+/// The set of Digital Signature Algorithm's natively supported within this SDK.
```
angiejones, line 114:
```suggestion
/// The set of Digital Signature Algorithms natively supported within this SDK.
```

docs/API_DESIGN.md
```diff
@@ -0,0 +1,658 @@
...
+
+**Version** 0.1.0
+
+**[Custom DSL](./CUSTOM_DSL.md) Version**: 0.1.0
+
+- [Key Material](#key-material)
+  - [`Jwk`](#jwk)
+  - [`InMemoryKeyManager`](#inmemorykeymanager)
+- [Cryptography](#cryptography)
+  - [Digital Signature Algorithm's (DSA)](#digital-signature-algorithms-dsa)
```
angiejones, line 13:
```suggestion
  - [Digital Signature Algorithms (DSA)](#digital-signature-algorithms-dsa)
```

docs/API_DESIGN.md
```diff
@@ -0,0 +1,658 @@
...
+```
+
+#### Example: Resolve a `did:dht`
+
+```pseudocode!
+uri = "did:dht:i9xkp8ddcbcg8jwq54ox699wuzxyifsqx4jru45zodqu453ksz6y"
+resolution_result = DidDht.resolve(uri)
+```
+
+# Verifiable Credential's (VCs)
```
decentralgabe, line 533:
```suggestion
# Verifiable Credentials (VCs)
```

decentralgabe (approved):
reviewed again and approving. this is an excellent starting point from which we can build further layers of abstraction.

docs/API_DESIGN.md
```diff
@@ -0,0 +1,658 @@
...
+  PUBLIC DATA fragment: string?
+
+  CONSTRUCTOR(uri: string)
+```
+
+### Example: Instantiate from a `did:dht`
+
+```pseudocode!
+uri = "did:dht:i9xkp8ddcbcg8jwq54ox699wuzxyifsqx4jru45zodqu453ksz6y"
+did = new Did(uri)
```
jiyoontbd, line 216: what can i do with this `Did`? usually i create a bearerdid so i can sign over something (i.e. a message)

jiyoontbd: it just turns the did string to an object?

docs/API_DESIGN.md
```diff
@@ -0,0 +1,658 @@
...
+  CONSTRUCTOR(public_key: Jwk)
+  CONSTRUCTOR(uri: string)
+  STATIC METHOD resolve(uri: string): ResolutionResult
+```
+
+#### Example: Create a `did:jwk`
+
+```pseudocode!
+key_manager = new InMemoryKeyManager()
+public_key = key_manager.generate_key_material()
```
jiyoontbd, line 417: can you tell me a bit more about what this function does? would an equivalent in the current impl be `KeyManager#GeneratePrivateKey()` which returns the keyid after storing the private key in the keystore (but now you're returning the public key as the keyid equivalent) ?

docs/API_DESIGN.md
```diff
@@ -0,0 +1,658 @@
...
+did_jwk = new DidJwk(public_key)
+```
+
+#### Example: Instantiate an existing `did:jwk`
+
+```pseudocode!
+uri = "did:jwk:eyJrdHkiOiJFQyIsInVzZSI6InNpZyIsImNydiI6InNlY3AyNTZrMSIsImtpZCI6ImkzU1BSQnRKS292SEZzQmFxTTkydGk2eFFDSkxYM0U3WUNld2lIVjJDU2ciLCJ4IjoidmRyYnoyRU96dmJMRFZfLWtMNGVKdDdWS04VEZaTm1BOVlnV3p2aGg3VSIsInkiOiJWTEZxUU1aUF9Bc3B1Y1hvV1gyLWJHWHBBTzFmUTVMbjE5VjVSQXhyZZVIiwiYWxnIjoiRVMyNTZLIn0"
+did_jwk = new DidJwk(uri)
```
jiyoontbd, line 425: this is only possible because it's a DidJwk, right?

docs/API_DESIGN.md
```diff
@@ -0,0 +1,658 @@
...
+```
+
+### `DidDht`
+
+```pseudocode!
+CLASS DidDht
+  PUBLIC DATA did: Did
+  PUBLIC DATA document: Document
+  CONSTRUCTOR(identity_key: Jwk)
+  CONSTRUCTOR(uri: string)
```
jiyoontbd, line 464: how could i create a DidDht with knowing just the uri?

docs/API_DESIGN.md
```diff
@@ -0,0 +1,658 @@
...
+> The call to the `new DidDht()` constructor only adds the minimum requirements to the DID Document.
+> If additional [Core Properties](https://www.w3.org/TR/did-core/#core-properties) are required, update the `document` data member prior-to the call to `publish()`.
+
+```pseudocode!
+key_manager = new InMemoryKeyManager()
+public_key = key_manager.generate_key_material()
+did_dht = new DidDht(public_key)
+
+/// Set the alsoKnownAs
+did_dht.document.alsoKnownAs = "did:example:efgh"
```
jiyoontbd, line 495: oof. this is not easy to do in all languages (we've already run into this issue in kotlin)

docs/API_DESIGN.md
jiyoontbd: central to what i want to ask here is 

> what can i _do_ with a DidWhateverMethod once instantiated?

docs/API_DESIGN.md
```diff
@@ -0,0 +1,658 @@
...
+  /// Returns the Ed25519Signer for the given public key.
+  METHOD get_signer(public_key: Jwk): Ed25519Signer
+
+  /// For importing keys which may be stored somewhere such as environment variables. Return Jwk is the public key for the given private key.
+  METHOD import_key(private_key: Jwk): Jwk
+```
+
+# Cryptography
+
+## Digital Signature Algorithms (DSA)
```
jiyoontbd, line 109: i see a lot of same concepts here as current impl, but can you tell me how the hierarchy changes with this design?

for example this is how i understand the current impl

image

docs/API_DESIGN.md
```diff
@@ -0,0 +1,658 @@
...
+  PUBLIC DATA id: string
+  PUBLIC DATA type: []string
+  PUBLIC DATA issuer: string | NamedIssuer
+  PUBLIC DATA issuanceDate: string
+  PUBLIC DATA expirationDate: string?
+  PUBLIC DATA credentialSubject: Object  # 🚧 `Object` not supported 🚧
+  CONSTRUCTOR(context: []string, id: string, type: []string, issuer: string | NamedIssuer, issuanceDate: string, expirationDate: string?)
+  CONSTRUCTOR(vcjwt: string)
+  CONSTRUCTOR(vcjwt: string, verifier: Verifier)
+  METHOD sign(signer: Signer): string
```
jiyoontbd, line 565: are you missing a `STATIC METHOD verify(vcjwt: string)` here?

KendallWeihe: `CONSTRUCTOR(vcjwt: string)` executes cryptographic verification. We're in kind of a halfsy-state wherein I have some `STATIC METHOD`'s but I'm slightly skeptical to full-commit to the concept of a static method because the concept varies from language to language and IMO we misuse the concept in a number of our implementations.

docs/API_DESIGN.md
KendallWeihe: The main design choice is to enable features which require cryptographic signing & verification to support both usage of a polymorphic base class called `KeyManager` (& actually in practice, it's a `BearerDid`) but also support direct usage of a DSA implementation by an implementation of the polymorphic base classes for `dsa::Signer` and `dsa::Verifier`. Point being, I can envision plenty of use cases, particularly large-scale enterprise environments, wherein the concept of a `KeyManager` would be considered excessively abstract and instead they need to bring-their-own cryptography through the usage of the DSA polymorphism classes.

This RFC is, of course, slightly out dated at this point, but that was a big part of the original intention.

docs/API_DESIGN.md
```diff
@@ -0,0 +1,658 @@
...
+
+### `Dsa`
+
+```pseudocode!
+/// The set of Digital Signature Algorithms natively supported within this SDK.
+ENUM Dsa
+  Ed25519
+```
+
+> We must add support for `Xd25519`, `secp256k1`, and `secp256r1` for [full did:dht conformance](https://did-dht.com/registry/index.html#key-type-index).
```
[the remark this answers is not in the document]
KendallWeihe: Oops, that's a typo heh -- it's `X25519`

https://www.rfc-editor.org/rfc/rfc7748#section-5

docs/API_DESIGN.md
KendallWeihe: You're referring to instantiating an instance and subsequently mutating it? What do you mean by "easy?"

docs/API_DESIGN.md
KendallWeihe: The `DidDht` is an encapsulation around the `Did` and `Document` abstraction, both of those are accessible via the DID URI. Admittedly I'm not sure this has a concrete use case right now 🤔

KendallWeihe: > in my mind, DIDs (perhaps it's more BearerDids) serve one purpose - to be a vessel that identifies me and holds a signing key so i can prove to bob that a thing i sent him actually came from me

Yeah you're referring specifically to bearer DID's, which is our own concept. The concept of a DID is actually moreso on the matter of accessing public keys, so for example if one wanted to verify a cryptographic signature and they had a DID URI, they could call this to gain access to the public key for verification.

docs/API_DESIGN.md
KendallWeihe: Yeah it's functionally equivalent

This was subsequently removed in favor of:

```pseudocode!
let key_manager = new InMemoryKeyManager();
let public_jwk = key_manager.import_private_jwk(Ed25519Generator::generate())
```

KendallWeihe: And yeah it's potay-to potah-to with regards to a string representing the thumbprint of the public key; both are functionally equivalent. I chose to go with the JWK being the point of reference because it's one less abstraction to accidentally break.

docs/API_DESIGN.md
KendallWeihe: this parses the DID URI into an object representing the Identifier from [the DID Core spec](https://www.w3.org/TR/did-core/#identifier), so it's use cases would be, for example, dereferencing wherein a framgment is needed to be parsed out. It's functionally equivalent to the concept of a `Url.parse(http_endpoint)` which many languages natively support, but specific to DID URI's

PR #230, opened 2024-06-12
fix: disable jsonschema resolution to support web5-rs in WASM builds

enmand:
Without disabling the resolve-http feature, the jsonschema crate fails to build on WASM environments due to use of `reqwest::blocking`, which is not built in WASM environments. Stranger6667/jsonschema-rs#222 has additional context.

The use of jsonschema, currently at least, does not use resolution and is used against a schema defined in code for Verifiable Credentials.


KendallWeihe (approved):
WASM [isn't within scope](https://github.com/TBD54566975/technical-steering-committee/pull/11#discussion_r1633247715) of the existing Rust Core Working Group, but so long as this doesn't impact the set of deliverables then it's great if we can enable WASM compatibility! ✅

PR #234, opened 2024-06-14
Implement partial APID 0.1.0 with UDL bindings

KendallWeihe:
This PR implements partial APID v0.1.0 proposed in #229 with comprehensive UDL bindings. This is not considered to be a complete implementation, but implements enough to unblock downstream Kotlin binded implementation (cc @nitro-neal ). The changes are compartmentalized to a new `web5::apid` module, and does not remove any of the prior `web5` code; the prior `web5` code still has code for reference (namely, VC, did:dht and did:web) which needs to be subsequently moved into this `apid` module, whereafter everything in the `apid` module will be moved into the root `web5` module.

## Missing Pieces

- Doc comments
- Tests & test vectors
- JOSE implementations using `josekit`
- did:dht

All of the above implementation will be implemented in subsequent pull requests.

## Known Limitations

- presentation definition's recursive `Filter`'s `contains` property is problematic, temporarily removed support 
- VC issuer only string support (temporary)

Both of the above limitations will be solvable, but may require additional development on the Kotlin side of the fence, so for the sake of agility I am temporarily removing support as to unblock kt development.


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [workspace]
 members = [ 
     "crates/web5", 
-    "bindings/uniffi",
+    "bind/uniffi",
```
nitro-neal, line 4: Hmm almost liking bindings here more actually, but I wonder if there is an official 'rust' way to do it

bind/uniffi/src/credentials/mod.rs
```diff
@@ -0,0 +1,2 @@
+pub mod presentation_definition;
+pub mod verifiable_credential_11;
```
nitro-neal, line 2: alternatively 
pub mod verifiable_credential_v1_1;

bind/uniffi/src/errors.rs
```diff
@@ -0,0 +1,107 @@
...
+pub enum RcbError {
+    #[error("{message}")]
+    Error {
+        message: String,
+        error_type: String,
+        error_variant: String,
+    },
+}
+
+impl RcbError {
```
nitro-neal, line 21: I dig it, this is a wrapper to all the other errors right?

With the kotlin code be able to determine it's type right by doing type_of?

bind/uniffi/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+            Constraints as RcbConstraintsData, Field as RcbFieldData, Filter as RcbFilterData,
+            InputDescriptor as RcbInputDescriptorData, Optionality as RcbOptionality,
+            PresentationDefinition as RcbPresentationDefinitionData,
+        },
+        verifiable_credential_11::{
+            CredentialSubject as RcbCredentialSubjectData,
+            VerifiableCredential as RcbVerifiableCredentialData,
+        },
+    },
+    dids::{
```
nitro-neal, line 42: ahh I thought the top level was going to be exported as RcbDids, to clear up the namespace in Kotlin.. but.. we can handle it at the kotlin layer :hmm:

bind/uniffi/src/web5.udl
```diff
@@ -0,0 +1,255 @@
...
+  string url;
+  string method;
+  string id;
+  record<string, string>? params;
+  string? path;
+  string? query;
+  string? fragment;
+};
+
+interface RcbDid {
```
nitro-neal, line 76: or maybe the kotlin will see RcbDid instead of "Did" but yea either can work!

crates/web5/src/apid/credentials/verifiable_credential_11.rs
```diff
@@ -0,0 +1,123 @@
...
+#[derive(Serialize, Deserialize, Debug, Clone, Default)]
+pub struct VerifiableCredential {
+    #[serde(rename = "@context")]
+    pub context: Vec<String>,
+    pub id: String,
+    #[serde(rename = "type")]
+    pub r#type: Vec<String>,
+
+    // 🚧 UDL support
+    // pub issuer: Issuer,
```
nitro-neal, line 54: if it stays a problem, we can make issuer a string only for now.
maybe need to get a issue for this.

Otherwise I will update it along with the josekit changes

nitro-neal: Oh perfect I see Issuer is string as of now, lol nice

nitro-neal (approved):
Legendary PR 👍 👍

Cargo.toml
KendallWeihe: Yeah I'm with you, switched back to `bindings/`

bind/uniffi/src/credentials/mod.rs
KendallWeihe: nice I like this too, done

bind/uniffi/src/errors.rs
KendallWeihe: yup! went with `RustCoreError` instead

KendallWeihe: also notice I slightly changed `error_variant()` to just `variant()` 

I wanted to do the same with `error_type()` but I'm concerned `type()` being a commonly used special word in programming languages that would cause issues

crates/web5/src/apid/jwk.rs
```diff
@@ -0,0 +1,50 @@
...
+#[derive(thiserror::Error, Debug, Clone, PartialEq)]
+pub enum JwkError {
+    #[error("thumbprint computation failed {0}")]
+    ThumbprintFailed(String),
+}
+
+type Result<T> = std::result::Result<T, JwkError>;
+
+impl Jwk {
+    pub fn compute_thumbprint(&self) -> Result<String> {
```
KendallWeihe, line 24: `compute_thumbprint()` isn't in the APID, but it's here as an implementation detail

over time we should consider adding it to the APID (but keeping lean for now)

PR #236, opened 2024-06-17
added bindings in reference to apid

nitro-neal:
# web5-core-kotlin

This contains the web5 Kotlin SDK that uses bindings from web5-rs.

As far as the API goes for Kotlin, this PR includes the following:

- Key Material
  - [x] `Jwk`
  - [x] `InMemoryKeyManager`
- Cryptography
  - Digital Signature Algorithms (DSA)
    - [ ] `Dsa`
    - [x] `Signer`
    - [ ] `Verifier`
    - [ ] `Ed25519Generator`
    - [x] `Ed25519Signer` (error, but debugging)
    - [ ] `Ed25519Verifier`
- Decentralized Identifiers (DIDs)
  - [x] `Did`
  - [x] `Document`
    - [x] `VerificationMethod`
    - [x] `Service`
  - [x] `ResolutionResult`
    - [x] `ResolutionMetadataError`
    - [x] `ResolutionMetadata`
    - [x] `DocumentMetadata`
  - Methods
    - [x] `DidJwk`
    - [ ] `DidWeb`
    - [x] `DidDht` (partial)

This has also created the unit test scaffolding and basic documentation


diehuxx (approved):

bound/kt/src/main/kotlin/web5/sdk/crypto/keys/InMemoryKeyManager.kt
```diff
@@ -0,0 +1,43 @@
...
+    private val rustCoreKeyManager = RustCoreInMemoryKeyManager()
+
+    /**
+     * Generates new key material and returns the public key represented as a Jwk.
+     *
+     * @return Jwk the public key represented as a Jwk.
+     */
+    override fun generateKeyMaterial(): Jwk {
+        val jwkData = rustCoreKeyManager.generateKeyMaterial()
+        return Jwk.fromBinding(jwkData)
```
KendallWeihe, line 19: Maybe `fromRustCore()` as to be consistent with the `as RustCoreInMemoryKeyManager`?

bound/kt/src/main/kotlin/web5/sdk/crypto/keys/InMemoryKeyManager.kt
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    /**
+     * Returns the Ed25519Signer for the given public key.
+     *
+     * @param publicKey the public key represented as a Jwk.
+     * @return Ed25519Signer the signer for the given public key.
+     */
+    override fun getSigner(publicKey: Jwk): Ed25519Signer {
+        return Ed25519Signer(publicKey)
```
KendallWeihe, line 29: Ah, the signer needs the private key so instead you'll need to make a call to `rustCoreKeyManager.getSigner()` with the public key (which you'll have to first convert to rust core JWK), and then you'll have to implement a `Ed25519Signer.fromBinding(rustCoreSigner)` (or like my prior comment recommended `Ed25519.fromRustCore(rustCoreSigner)`)

So yeah, the `Ed25519Verifier` takes a public key in its constructor and the signer takes a private key

KendallWeihe: So yeah, something like this

```suggestion
        val rustCorePublicKey = publicKey.toRustCore()
        val rustCoreSigner = rustCoreKeyManager.getSigner(rustCorePublicKey)
        return Ed25519Signer.fromRustCore(rustCoreSigner)
```

bound/kt/src/main/kotlin/web5/sdk/crypto/keys/Jwk.kt
```diff
@@ -0,0 +1,71 @@
+package web5.sdk.crypto.keys
+
+import web5.sdk.rust.JwkData as RustCoreJwkData
+
+/**
+ * Partial representation of a [JSON Web Key as per RFC7517](https://tools.ietf.org/html/rfc7517).
+ * Note that this is a subset of the spec.
+ */
+data class Jwk(
```
KendallWeihe, line 9: since `RustCoreJwkData` is solely a data structure, without any instance methods (or another way of saying this is, since `Jwk` in the APID 0.1.0 is solely a data structure without instance or static methods or constructors)... I wonder if you could just re-export `import web5.sdk.rust.JwkData as RustCoreJwkData` instead of rewriting the data class? 🤔 trying to exercise my kt muscles... if yes, then the same would hold true for any type in the APID which is solely a data structure, such as `Document` or `VerificationMethod` or `Service` or `DocumentMetadata` or `InputDescriptor` and so on...

if so, it would save a lot of overhead, right? because conversions would be unnecessary

KendallWeihe: this may be unwise, so just an idea I'm throwing out for consideration

bound/kt/src/main/kotlin/web5/sdk/crypto/keys/KeyManager.kt
```diff
@@ -0,0 +1,32 @@
+package web5.sdk.crypto.keys
+
+import web5.sdk.crypto.signers.Signer
+
+/**
+ * An interface representing a key manager for cryptographic operations.
+ */
+interface KeyManager {
```
KendallWeihe, line 8: this isn't in the APID 0.1.0 FWIW but will probably come in the next version when we add bearer DIDs. it's no problem to include it here, but just calling it out in case you weren't aware (& also it may change when we add it in the next version of the APID). it's legal, according to RFC-0001 for implementations to go above the APID, so this is allowed per the discretion of each implementor

> The APID is asserted to act as a floor, or a baseline, of feature support — each implementation may choose to expose publicly facing APIs which are beyond the scope of the APID.

bound/kt/src/main/kotlin/web5/sdk/crypto/signers/RustSignerAdapter.kt
```diff
@@ -0,0 +1,11 @@
+package web5.sdk.crypto.signers
+
+import web5.sdk.rust.SignerInterface as RustSignerInterface
+
+// TODO: Figure out if we need this
+class RustSignerAdapter(private val kotlinSigner: Signer) : RustSignerInterface {
```
KendallWeihe, line 6: is this because of something weird UniFFI is code-gen'ing? not a show stopper just highlighting it in case you glossed over it prior to requesting review

bound/kt/src/main/kotlin/web5/sdk/crypto/keys/Jwk.kt
KendallWeihe: so but yeah to be clear, if we have a `CLASS` in the APID which has anything other than a set of `PUBLIC DATA` then we'll definitely have to wrap the rust core bound code b/c UDL interfaces don't conform to the APID

bound/kt/src/main/kotlin/web5/sdk/dids/DidResolutionResult.kt
```diff
@@ -0,0 +1,144 @@
...
+ * Representation of the result of a DID (Decentralized Identifier) resolution.
+ *
+ * @property didDocument The resolved DID document, if available.
+ * @property didDocumentMetadata The metadata associated with the DID document.
+ * @property didResolutionMetadata The metadata associated with the DID resolution process.
+ */
+data class DidResolutionResult(
+    val didDocument: DidDocument? = null,
+    val didDocumentMetadata: DidDocumentMetadata? = null,
+    val didResolutionMetadata: DidResolutionMetadata = DidResolutionMetadata()
```
KendallWeihe, line 18: technically in the APID the `did*` prefix on these data member names is not included, so just `document` or `documentMetadata` https://github.com/TBD54566975/web5-rs/blob/kendall/rfc-0001/docs/API_DESIGN.md#resolutionresult

I do think we'll need consistency in the name, albeit different casing, in order to be considered conformant to the APID. we could change the APID tho, I'm not opposed

KendallWeihe (approved):
I left a handful of comments I hope are helpful but we can go ahead and get this into main whenever you're ready for it @nitro-neal

bound/kt/src/main/kotlin/web5/sdk/crypto/signers/RustSignerAdapter.kt
nitro-neal: ^ Yea I was going to give it another shot today, I think I need something like this to do diddht.publish()

bound/kt/src/main/kotlin/web5/sdk/dids/DidResolutionResult.kt
nitro-neal: Yea I could go either way, I noticed it was just "Document" Seemed really generic, but I'll stick with it! 

I'll drop the Did on these to go towards the APID for now, and we can chnage later if we change APID

bound/kt/src/main/kotlin/web5/sdk/crypto/keys/InMemoryKeyManager.kt
nitro-neal: Yup lets do it ✅

PR #238, opened 2024-06-18
Add typealias and update udl signer to Ed25519Signer

nitro-neal:
Add typealias and update udl signer to Ed25519Signer

All data classes are now typealiased. I think this will make our development much faster, if we can use it in all circumstances I think we should use it until we can't.

This change also fixes the getSigner method and fixed the udl code to correctly have the method signature with `Ed25519Signer`


KendallWeihe:
> we should use it until we can't

Agreed 👍

KendallWeihe (approved):
let's go ⚡ fast turn around!

PR #240, opened 2024-06-20
RFC-0002 Basic `BearerDid`

KendallWeihe:
🚧


docs/API_DESIGN.md
```diff
@@ -530,6 +548,20 @@ uri = "did:dht:i9xkp8ddcbcg8jwq54ox699wuzxyifsqx4jru45zodqu453ksz6y"
 resolution_result = DidDht.resolve(uri)
 ```
 
+## `BearerDid`
+
+```pseudocode!
+CLASS BearerDid
+  PUBLIC DATA did: Did
+  PUBLIC DATA document: Document
+  CONSTRUCTOR(did: Did, key_manager: KeyManager)
```
KendallWeihe, line 557:
```suggestion
  CONSTRUCTOR(uri: string, key_manager: KeyManager)
```

For consistency, for now. IMO we should migrate every instance of `uri: string` to `did: Did`, that way the errors are pushed closer to the developer in the case the DID URI is invalid

docs/API_DESIGN.md
```diff
@@ -2,14 +2,15 @@
 
 **Last Updated** May 30, 2024
 
-**Version** 0.1.0
```
nitro-neal, line 5: sure you dont want to save 1.0.0 for the for sure for sure globalish release?

KendallWeihe: hehe I was hoping someone would see this

no

🙂 I want to set aside marketing material and _**BUILD**_

docs/API_DESIGN.md
```diff
@@ -89,23 +93,30 @@ CLASS Jwk
...
+  CONSTRUCTOR(private_jwks: []Jwk)
 
-  /// Returns the Ed25519Signer for the given public key.
-  METHOD get_signer(public_key: Jwk): Ed25519Signer
+  /// Returns the signer for the given public key.
+  METHOD get_signer(public_jwk: Jwk): Signer
 
   /// For importing keys which may be stored somewhere such as environment variables. Return Jwk is the public key for the given private key.
-  METHOD import_key(private_key: Jwk): Jwk
+  METHOD import_private_jwk(private_jwk: Jwk): Jwk
```
nitro-neal, line 117: Yea its not 100% clear from a new users prospective  (unless they read the documentation) that the return JWK will be public instead of private.

So this seems like another datapoint to have a light wrapper around Jwk for PrivateJwk and PublicJwk

docs/API_DESIGN.md
```diff
@@ -517,7 +535,7 @@ did_dht.document.alsoKnownAs = "did:example:efgh"
 /// Note: you could also add a verification method, set the controller etc.
 
 key_manager = new InMemoryKeyManager()
-public_key = key_manager.import_key(private_key) /// assume private_key pre-exists, eg. read from env var
+public_key = key_manager.import_private_jwk(private_key) /// assume private_key pre-exists, eg. read from env var
```
nitro-neal, line 538: 👍

nitro-neal (approved):

docs/API_DESIGN.md
```diff
@@ -0,0 +1,690 @@
...
+
+### `KeyManager`
+
+```pseudocode!
+INTERFACE KeyManager
+  /// Returns the signer for the given public key.
+  METHOD get_signer(public_jwk: Jwk): Signer
+```
+
+### `InMemoryKeyManager`
```
decentralgabe, line 106: can we clarify that this is to be used for test purposes only

docs/API_DESIGN.md
jiyoontbd: is that perhaps the reason why the current implementation would return keyId string instead of the public key, after storing the private key?

docs/API_DESIGN.md
KendallWeihe: done

PR #241, opened 2024-06-20
Implement namespacing from RFC-0002

KendallWeihe:
Simply moving code around, nothing new or added, just different namespaces in accordance with #240


KendallWeihe:
Closing in favor of #242 

I thought opening this PR would assist with review, but I now see that was futile

PR #242, opened 2024-06-20
Implement RFC-0002 BearerDid

KendallWeihe:
🚧

Implementation for #240 for rust and bindings, ready for kt binding


bindings/uniffi/Cargo.toml
```diff
@@ -10,6 +10,7 @@ license-file.workspace = true
 thiserror = { workspace = true }
 web5 = { path = "../../crates/web5" }
 uniffi = { version = "0.27.1", features = ["cli"] }
+web5_uniffi_wrapper = { path = "../uniffi-wrapper" }
```
nitro-neal, line 13: I noticed you have an `_` above for the package name but a `-` for the file name

crates/web5/src/apid/crypto/key_managers/in_memory_key_manager.rs
```diff
@@ -1,60 +1,55 @@
...
+}
+
 impl InMemoryKeyManager {
     pub fn new() -> Self {
         Self {
             map: RwLock::new(HashMap::new()),
         }
     }
 
-    pub fn generate_key_material(&self) -> Result<Jwk> {
```
nitro-neal, line 33: so now we have to do Ed2219 ed25519_generator_generate and import private key?

nitro-neal (approved):
Awesome! I don't know why we have to use this uniffi-wrapper, I figure its for wrangling the data to be binded in a specific way that you can't do with just the one uniffi file

crates/web5/src/apid/crypto/key_managers/in_memory_key_manager.rs
nitro-neal: or just the old school way now, pass a key manager into bearer did

KendallWeihe: that's right, I added some details to the [Draft RFC PR](https://github.com/TBD54566975/web5-rs/pull/240/files)

> Removal of `InMemoryKeyManager`'s `METHOD generate_key_material()`: this is nonsensical since the key manager implementation stores keys in memory and does not expose the functionality to export keys. The only use case would be for short-lived ephermal use cases, in which case, the developer can use the `Ed25519Generator` to generate the keys and call `METHOD import_jwk()` or the new `CONSTRUCTOR(private_jwks: []Jwk)`.

We're moving fast to unblock development, but I'll still backfill reasoning in the RFCs

We'll probably still want to expose a generator function, but I realized I was getting ahead of myself with convenience things, so removing for now

nitro-neal:
This brings up a good point that we will need to bring up sooner or later.

-I think- that since this changed the bindings and regenerated `web5.kt` that when someone pulls web5-kt it will be broken out of the box.

I think we need some kind of internal versioning.

I imagine `just bind` would create the `web5.kt` file in a different directory outside the `kt` folder. Maybe just outside in a `latest` folder.

We can make another `just` command which is `just update-kt` or something which will just copy the `latest/web5.kt` into the official kt bound folder, and at that point it would break and the dev would have to update the kt to fix the bindings.

KendallWeihe:
> I don't know why we have to use this uniffi-wrapper, I figure its for wrangling the data to be binded in a specific way that you can't do with just the one uniffi file

That's right, and I think we can clean this up later, but for some reason I can't dig into right now, and I think it's a result of the custom UniFFI `build.rs` script, I can't consume the Rust code from the wrapper in the tbdex UniFFI crate without splitting it out into a dedicated crate. And yeah I've come to terms that we'll have to effectively wrap everything other than purely data structures, it's just how it be, but it's fine because it's a recipe to follow.

KendallWeihe:
> This brings up a good point that we will need to bring up sooner or later.\
> 
> -I think- that since this changed the bindings and regenerated web5.kt that when someone pulls web5-kt it will be broken out of the box.
> 
> I think we need some kind of internal versioning.

Great point, I shouldn't actually generate the kt bindings, I should hold off for you to do it when you make the kt changes. Thanks, I'll keep this in mind next time!

bindings/uniffi/Cargo.toml
KendallWeihe: Good catch, I don't like that inconsistency, I'll fix. Also I'm going to make the directory name match the exact rust crate name. So:

- `bindings/uniffi` --> `bindings/web5_uniffi`
- `bindings/uniffi-wrapper` --> `bindings/web5_uniffi_wrapper`

KendallWeihe: done

KendallWeihe:
> Great point, I shouldn't actually generate the kt bindings, I should hold off for you to do it when you make the kt changes. Thanks, I'll keep this in mind next time!

@nitro-neal actually I went ahead and restored `bound/` from `main` so this PR shouldn't break the existing kt project

PR #244, opened 2024-06-20
Fix Signer Implementation in Ed25519Signer and InMemoryKeyManager.getSigner

nitro-neal:
Fix Signer Implementation in Ed25519Signer and InMemoryKeyManager.getSigner

We are now exporting the Signer interface type directly instead of having to do some wrapping and conversion of kotlin signer and rust signer

The rust bindings are generated like this:
```
class Signer implements SignerInterface
```

our signer implements the sign function, so we use `SignerInterface`:
```
import web5.sdk.rust.SignerInterface as RustCoreSignerInterface

typealias Signer = RustCoreSignerInterface
```


KendallWeihe (approved):
eggsellent

PR #249, opened 2024-06-24
Add Presentation Definition Object

nitro-neal:
Add Presentation Definition Object

To create one it will look like this:
```
val inputDescriptor = InputDescriptor(
      id = "test_input",
      name = "Test Input",
      purpose = "For testing",
      constraints = ConstraintsData(
        fields = listOf(
          FieldData(
            id = "field1",
            name = "Field 1",
            path = listOf("$.field1"),
            purpose = "Test field",
            filter = FilterData(
              type = "string",
              pattern = "^[a-zA-Z]+$",
              constValue = null
            ),
            optional = false,
            predicate = Optionality.REQUIRED
          )
        )
      )
    )

    val presentationDefinition = PresentationDefinition(
      id = "test_presentation",
      name = "Test Presentation",
      purpose = "For testing purposes",
      inputDescriptors = listOf(inputDescriptor)
    )
```

### DIFF
--- a/bound/kt/src/main/kotlin/web5/sdk/vc/pex/PresentationDefinition.kt
+++ b/bound/kt/src/main/kotlin/web5/sdk/vc/pex/PresentationDefinition.kt
@@ -1,0 +1,49 @@
+package web5.sdk.vc.pex
+
+import web5.sdk.rust.ConstraintsData as RustCoreConstraintsData
+import web5.sdk.rust.FieldData as RustCoreFieldData
+import web5.sdk.rust.FilterData as RustCoreFilterData
+import web5.sdk.rust.Optionality as RustCoreOptionality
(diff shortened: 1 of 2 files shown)


bound/kt/src/main/kotlin/web5/sdk/vc/pex/PresentationDefinition.kt
```diff
@@ -0,0 +1,49 @@
...
+typealias Optionality = RustCoreOptionality
+typealias InputDescriptor = RustCoreInputDescriptorData
+
+class PresentationDefinition {
+    val id: String
+    val name: String?
+    val purpose: String?
+    val inputDescriptors: List<InputDescriptor>
+
+    internal val rustCorePresentationDefinition: RustCorePresentationDefinition
```
KendallWeihe, line 24: nice! `internal`

KendallWeihe (approved):
🎉

PR #250, opened 2024-06-24
Create, Publish, and Resolve did:dht

diehuxx:


crates/web5/src/apid/dids/methods/did_dht/mod.rs
```diff
@@ -28,19 +28,8 @@ pub mod document_packet;
 const JSON_WEB_KEY: &str = "JsonWebKey";
 const DEFAULT_RELAY: &str = "https://diddht.tbddev.org";
 
-fn extract_ed25519_public_key(jwk: &Jwk) -> Result<Vec<u8>> {
-    let decoded_x = general_purpose::URL_SAFE_NO_PAD.decode(&jwk.x)?;
-    if decoded_x.len() != PUBLIC_KEY_LENGTH {
-        return Err(MethodError::DidCreationFailure(format!(
-            "Identity key JWK public key must have length {}",
```
KendallWeihe: I removed this in favor of `ed25519::extract_public_key`. We lose out on the slight difference in error handling, but I think it's fine in service of DRY.

crates/web5/Cargo.toml
```diff
@@ -33,8 +33,10 @@ ssi-jwk = "0.1.2"
 thiserror = { workspace = true }
 url = "2.5.0"
 uuid = { version = "1.8.0", features = ["v4"] }
+zbase32 = "0.1.2"
 
 [dev-dependencies]
 chrono = { workspace = true }
+mockito = "1.4.0"
```
KendallWeihe, line 40: did we end up not using this? all good if not, but we should remove

KendallWeihe: done

crates/web5/src/apid/dids/methods/did_dht/mod.rs
```diff
@@ -44,14 +115,117 @@ impl DidDht {
     }
 
     pub fn resolve(uri: &str) -> Result<ResolutionResult> {
```
KendallWeihe, line 117: I appreciate the staying-true to Rust idioms with the return value of `Result<ResolutionResult>` here, but to be conformant to the spec, we should return `ResolutionResult` and use the embedded `ResolutionMetadataError` for error cases (those which are specific to rust runtime errors such as json serialization, etc. would be `ResolutionMetadataError::InternalError`. It's a bit awkward, because we follow the monad pattern with `Result<T>` everywhere, but yeah we should be conformant to the spec and utilize the embedded `error` data member. 

You can see how I did this over in `did:jwk` using a closure (admittedly it may not be exactly the same here because we still want to capture the `ResolutionMetadataError` variants in the places you're early-returning via the `?` -- as opposed to in `did:jwk` it all maps to the `InternalError` variant.

I can take a stab at this @diehuxx and you can review and lmk if you think I did things incorrectly.

KendallWeihe: Okay done here, lmk if you take issue @diehuxx https://github.com/TBD54566975/web5-rs/pull/250/commits/dff3fa72927ecc44868a49d89dd5ded88b2c27ec

Also it made me realize I needed to improve the error cases over in the did:jwk resolve()

KendallWeihe: Also realizing I need to fix did:web accordingly too, but will do that in a different PR

crates/web5/src/apid/dsa/ed25519.rs
```diff
@@ -28,6 +28,45 @@ impl Ed25519Generator {
...
+    Ok(Jwk {
+        alg: "EdDSA".to_string(),
+        kty: "OKP".to_string(),
+        crv: "Ed25519".to_string(),
+        x: general_purpose::URL_SAFE_NO_PAD.encode(public_key),
+        ..Default::default()
+    })
+}
+
+#[cfg(test)]
```
KendallWeihe, line 48: fascinating, TIL you can do this for tests only, nice!

crates/web5/src/apid/dsa/ed25519.rs
```diff
@@ -28,6 +28,45 @@ impl Ed25519Generator {
     }
 }
 
+pub(crate) fn from_public_key(public_key: &[u8]) -> Result<Jwk> {
```
KendallWeihe, line 31: I really appreciate this function and the `extract_public_key` function below, and I really appreciate how they're both `pub(crate)`! and I also appreciate it's inside the `ed25519` module! very thoughtful!

but I think we should call them `public_jwk_from_bytes()` and `public_jwk_extract_bytes()`; this way we enable ourselves down the lines to support more key representations and more primitive types (other than byte arrays) for the input/output. so you could imagine `public_multibase_from_string()`

crates/web5/src/apid/dsa/mod.rs
```diff
@@ -1,4 +1,5 @@
 pub mod ed25519;
+pub(crate) mod secp256k1;
```
KendallWeihe, line 2: nice!!! I love how we added it, but we aren't committing (yet) to public support (but we will! with more thoughtful consideration), thank you!

KendallWeihe (approved):

PR #253, opened 2024-06-27
Move all apid code to root module

KendallWeihe:
Mostly moving code around and updating dependency paths (some of the git diff views look like code changes but that's a result of what was inside the `apid` module was largely a copy-paste of the old™️ code)

Had to temporarily disable test vectors in `presentation_definition.rs` because the `web5-spec` test vectors use `did:key` which we don't support (previously working because we have custom written `CompactJws` which decoded without verifying, but now we're verifying the VC-JWT) -- this is fine, test vectors are a next big priority.


crates/web5/src/credentials/presentation_definition.rs
```diff
@@ -71,23 +70,25 @@ pub struct Field {
...
     #[serde(skip_serializing_if = "Option::is_none")]
     pub pattern: Option<String>,
     #[serde(rename = "const", skip_serializing_if = "Option::is_none")]
     pub const_value: Option<String>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub contains: Option<Box<Filter>>,
+    // 🚧 UDL doesn't support nested types -- need to find an alternative, for now removing support
+    // 🚧 https://github.com/TBD54566975/web5-rs/pull/234
+    // #[serde(skip_serializing_if = "Option::is_none")]
+    // pub contains: Option<Box<Filter>>,
```
KendallWeihe, line 91: this is a next big priority

crates/web5/src/credentials/presentation_definition.rs
```diff
@@ -172,11 +173,10 @@ impl InputDescriptor {
         for vc_jwt in vc_jwts {
             let mut selection_candidate: Map<String, Value> = Map::new();
 
-            let decoded_jws = CompactJws::decode(vc_jwt)?;
-            let payload_json = String::from_utf8(decoded_jws.payload).map_err(|_| {
-                PexError::JsonError(
-                    "Could not create json string from vc jwt payload bytes".to_string(),
-                )
+            let vc = VerifiableCredential::verify(vc_jwt)
```
KendallWeihe, line 176: this is what's making the test vectors fail now, because `did:key` isn't supported here, but the test vectors use VCs created with `did:key`'s

KendallWeihe: Actually we should slightly adjust this to use a continue statement instead of failing. The algorithm here should account for failed verification among the set of VCs and still be passable, because the intent of the function isn't to verifying everything but to select matching credentials, and for the given matches we should verify.

crates/web5/src/credentials/presentation_definition.rs
KendallWeihe: done

bound/kt/src/main/kotlin/web5/sdk/Json.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import com.fasterxml.jackson.annotation.JsonInclude
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.databind.ObjectWriter
+import com.fasterxml.jackson.databind.SerializationFeature
+import com.fasterxml.jackson.module.kotlin.registerKotlinModule
+
+internal const val dateTimeFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSXXX"
+
+internal object Json {
```
nitro-neal, line 11: need to look at the difference a little more, but a lot of times my ide gives me a yellow suggestion that I should make internal private.

Also should we rename this Web5Json or something, to lower confusion ?

bound/kt/src/main/kotlin/web5/sdk/vc/pex/PresentationDefinition.kt
```diff
@@ -1,49 +1,54 @@
...
+    )
 
     fun selectCredentials(vcJwts: List<String>): List<String> {
         return this.rustCorePresentationDefinition.selectCredentials(vcJwts)
     }
-}
\ No newline at end of file
+}
+
+data class InputDescriptor(
```
nitro-neal, line 23: yea I was going back and fourth with this, but seems now we don't have to have toRustCore / fromJustCore now? 

We could `typealias InputDescriptor = RustCoreInputDescriptorData` instead, but It is definitely cleaner (especially if users are going to be creating these types which they wyill) to redefine it like you did here.

So overall, are we back on the train of not doing typealias and just redefining most things since things got updated?

crates/web5/src/dids/methods/did_dht/mod.rs
```diff
@@ -26,6 +28,24 @@ pub mod document_packet;
...
+}
+
+impl std::str::FromStr for KeyType {
+    type Err = DsaError;
+
+    fn from_str(input: &str) -> std::result::Result<Self, DsaError> {
+        match input.to_ascii_lowercase().as_str() {
+            "ed25519" => Ok(KeyType::Ed25519),
+            "secp256k1" => Ok(KeyType::Secp256k1),
+            _ => Err(DsaError::UnsupportedCurve),
```
nitro-neal, line 44: Did we rebrand to `UnsupportedKeyType`

crates/web5/src/dids/resolution/resolution_result.rs
```diff
@@ -31,6 +31,19 @@ impl ResolutionResult {
         match did.method.as_str() {
             "jwk" => DidJwk::resolve(uri),
             "dht" => DidDht::resolve(uri),
+            "web" => {
```
nitro-neal, line 34: why can't this be DidWeb::resolve ?

nitro-neal: or maybe make internal DidWeb::resolve do all this thread stuff

crates/web5/src/dids/resolution/resolution_result.rs
```diff
@@ -53,4 +66,25 @@ mod tests {
         assert_eq!(None, resolution_result.resolution_metadata.error);
         assert_eq!(resolution_result.document.unwrap().id, did_uri.to_string());
     }
+
+    #[test]
+    fn can_resolve_did_web() {
+        let did_uri = "did:web:tbd.website";
+        let resolution_result = ResolutionResult::new(did_uri);
+
+        // the did:web we host is currently invalid hehe https://www.tbd.website/.well-known/did.json
```
nitro-neal, line 75: yikez

bound/kt/src/main/kotlin/web5/sdk/Json.kt
KendallWeihe: I'm open to whatever works best so long as we're not exposing this as a publicly facing api

crates/web5/src/dids/resolution/resolution_result.rs
nitro-neal: there is another endpoint we have on a slightly different path

bound/kt/src/main/kotlin/web5/sdk/vc/pex/PresentationDefinition.kt
KendallWeihe: we can typealias types which don't require json serialization over the FFI; right now the constraints which lead us to json serialize over the FFI are:

- recursive types (types which have a data member which is itself the same type)
- types which have a "JSON Object" data members

neither of the above two conditions are supported at the FFI layer so we have to json serialize over the boundary, and so because of that, we have to re-define the types at the kt layer

nitro-neal (approved):
Lets go!

crates/web5/src/dids/resolution/resolution_result.rs
KendallWeihe: yeah nice, agreed. done!

Issue #128, opened 2024-04-24
Fully specify Ed25519 in JOSE features


KendallWeihe:
Originally brought up here https://github.com/TBD54566975/web5-rs/pull/97#discussion_r1576858213

We need to fully-specifcy Ed25519 rather than EdDSA in our algorithm values

I actually did went ahead and did this [here](https://github.com/TBD54566975/web5-rs/commit/460d8ba70bf8014c6dde13ff875cae79fdb02cfa) but in the process it seems like there are still some unknowns so we can dedicate time to do this right 

@decentralgabe should we still support verification of a JWS Header w/ Ed25519 given the alg value is EdDSA?

Also during the work I realized our did:jwk create function may have room for improvement, we're doing some weird things to use the underlying spruce lib which feels off. Also all of the verification relationships are empty and shouldn't be. Feel free to fix these things here or in a separate work item.

decentralgabe:
> we still support verification of a JWS Header w/ Ed25519 given the alg value is EdDSA?

yes, since what I mentioned is still in draft. the reason I support adopting it immediately is because it reduces confusion. it is sensible to support both as the transition takes place

frankhinek:
Agreed with @decentralgabe 's note above.  Since many libraries won't adopt the new `Ed25519` alg anytime soon we'll need to ensure we at least support both... even if we prefer the more description algorithm identifier.

PR #254, opened 2024-06-28
RFC-0003 `PortableDid` & `web5` CLI

KendallWeihe:
Add RFC-0003 which includes the addition of the `PortableDid`; please read the RFC for full context.

Implemented a basic `web5` CLI (which makes use of the `PortableDid` -- see the README for sample usage.

---

Also fixed a small bugs in `crates/web5/src/dids/methods/did_dht/bep44.rs`, `crates/web5/src/crypto/dsa/ed25519.rs`, and `crates/web5/src/dids/data_model/document.rs`.

Also fixed RFC-0002 which was accidentally left behind.


docs/API_DESIGN.md
```diff
@@ -682,9 +685,85 @@ CLASS BearerDid
...
 > [!WARNING]
 >
-> We'll need to add support for the developer to select a VM other than defaulting to the first in the `verification_method` array; add `METHOD get_signer_by_kid(key_id: string): Signer`.
\ No newline at end of file
+> We'll need to add support for the developer to select a VM other than defaulting to the first in the `verification_method` array; add `METHOD get_signer_by_kid(key_id: string): Signer`.
+
+### Example: Instantiate from a [`PortableDid`](#portabledid)
+
+```pseudocode!
+portable_did = JSON.parse(env.get("PORTABLE_DID"))
```
KendallWeihe, line 699: we have a `CONSTRUCTOR(json: string)` for good reason (to give more useful error results), so we should use it here!

KendallWeihe:
Need to merge #255 and #256 prior to merging this, but won't do so until review is completed because it will bloat the review of the changes here

Cargo.toml
```diff
@@ -1,8 +1,9 @@
 [workspace]
 members = [ 
+    "crates/cli",
```
nitro-neal, line 3: i wonder if we should name it web5_cli to keep the theme going, and then maybe one day there will be a tbdex_cli

crates/cli/src/dids/create.rs
```diff
@@ -0,0 +1,94 @@
...
+impl Commands {
+    pub fn command(&self) {
+        match self {
+            Commands::Jwk {
+                no_indent,
+                json_escape,
+            } => {
+                let private_jwk = Ed25519Generator::generate();
+                let mut public_jwk = private_jwk.clone();
+                public_jwk.d = None;
```
nitro-neal, line 54: yea these are strange, more data points that we need a public/private Jwk 'lite' type

crates/web5/src/crypto/dsa/ed25519.rs
```diff
@@ -37,7 +37,7 @@ pub(crate) fn public_jwk_from_bytes(public_key: &[u8]) -> Result<Jwk> {
     }
 
     Ok(Jwk {
-        alg: "EdDSA".to_string(),
+        alg: "Ed25519".to_string(),
```
nitro-neal, line 40: you sure this is right?
image

image

I'm a bit confused on this issue at this point, but pretty much all other sources have an ed25519 jwk the alg set to EdDSA

docs/API_DESIGN.md
```diff
@@ -682,9 +685,85 @@ CLASS BearerDid
...
+    "capabilityInvocation": [
+      "did:dht:4nca8jd5q5qwrowbx1efrihkac6danj6fpkhrrnrhdifiq19xfry#0"
+    ],
+    "capabilityDelegation": [
+      "did:dht:4nca8jd5q5qwrowbx1efrihkac6danj6fpkhrrnrhdifiq19xfry#0"
+    ]
+  },
+  "privateKeys": [
+    {
+      "alg": "Ed25519",
```
nitro-neal, line 761: if you update the above, reupdate these

web5
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+cargo build -p cli > /dev/null 2>&1
+cp ./target/debug/cli ./target/debug/web5
```
nitro-neal, line 4: should the path still be debug?

nitro-neal (approved):
few small things but overall ✅

crates/web5/src/crypto/dsa/ed25519.rs
KendallWeihe: yup we're moving to fully-specific alg https://github.com/TBD54566975/web5-rs/issues/128

Cargo.toml
KendallWeihe: agreed, done

PR #255, opened 2024-06-29
Add foreign trait implementation support

KendallWeihe:
Add support for Kotlin to implement it's own instance of `Signer`, `Verifier` or `KeyManager` and utilize it inside the Rust bindings. Accomplished via [the `[WithForeign]` feature from UniFFI](https://mozilla.github.io/uniffi-rs/0.27/udl/interfaces.html#foreign-implementations).

As an added benefit of this, reworking the `RustCoreError` automatically prints the `type`, `variant` and `msg` line the following:

```
web5.sdk.rust.RustCoreException$Exception: type=test-type, variant=test-varient, msg=test-msg
```

### DIFF
--- a/bindings/web5_uniffi/src/web5.udl
+++ b/bindings/web5_uniffi/src/web5.udl
@@ -193,6 +193,18 @@ interface DidDht {
   DidDhtData get_data();
 };
 
+dictionary PortableDidData {
+  string did_uri;
+  DocumentData document;
+  sequence<JwkData> private_jwks;
+};
+
+interface PortableDid {
+  [Throws=RustCoreError]
+  constructor([ByRef] string json);
+  PortableDidData get_data();
+};
+
 dictionary BearerDidData {
   DidData did;
   DocumentData document;
@@ -202,6 +214,8 @@ dictionary BearerDidData {
 interface BearerDid {
   [Throws=RustCoreError]
   constructor([ByRef] string uri, KeyManager key_manager);
+  [Throws=RustCoreError, Name=from_portable_did]
+  constructor(PortableDid portable_did);
   BearerDidData get_data();
   [Throws=RustCoreError]
   Signer get_signer(string key_id);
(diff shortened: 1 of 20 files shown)


bindings/web5_uniffi/src/web5.udl
```diff
@@ -42,28 +41,28 @@ enum Dsa {
   "Ed25519"
 };
 
-[Trait]
+[Trait, WithForeign]
 interface Signer {
   [Throws=RustCoreError]
-  bytes sign([ByRef] sequence<u8> payload);
+  bytes sign(bytes payload);
```
KendallWeihe, line 47: Oh nice, and @nitro-neal this improves things too, because no longer do we need to use `UByteArray` and instead can just use `ByteArray` (discovered this because [I'm reworking tbdex-rs](https://github.com/TBD54566975/tbdex-rs/pull/64) to also have these improvements)

KendallWeihe: (So when we run `just bind` next time, this will probably break things, because we'll have to rework the types)

nitro-neal: Awesome!

bindings/web5_uniffi_wrapper/src/crypto/dsa/mod.rs
```diff
@@ -5,23 +5,35 @@ use std::sync::Arc;
 use web5::crypto::dsa::{Signer as InnerSigner, Verifier as InnerVerifier};
 
 pub trait Signer: Send + Sync {
-    fn sign(&self, payload: &[u8]) -> Result<Vec<u8>>;
-    fn to_inner(&self) -> Arc<dyn InnerSigner>;
+    fn sign(&self, payload: Vec<u8>) -> Result<Vec<u8>>;
 }
 
-pub struct OuterSigner(pub Arc<dyn InnerSigner>);
+pub struct ToOuterSigner(pub Arc<dyn InnerSigner>);
```
nitro-neal, line 11: make sure to change everything else in the codebase to `ToOuter` from `Outer`

PR #259, opened 2024-07-07
Make Jwk.alg optional, add did:web localhost resolution support

KendallWeihe:


KendallWeihe:
This PR doesn't include binding changes, because we need to first implement build systems for multi-targets as we did in `tbdex-rs` here https://github.com/TBD54566975/tbdex-rs/pull/67

crates/web5/src/dids/methods/did_web/resolver.rs
```diff
@@ -34,7 +34,14 @@ impl Resolver {
         };
 
         Self {
-            did_url: format!("https://{}/did.json", did_url),
+            did_url: format!(
+                "{}://{}/did.json",
+                match did_url.contains("localhost") {
```
nitro-neal, line 39: should this be == localhost?
What if I have neal-localhost.com ?

nitro-neal (approved):
we need to `just bind` and fix kt after this right?

crates/web5/src/dids/methods/did_web/resolver.rs
KendallWeihe: yeah we need to use the `Url` crate, but will do subsequently

PR #263, opened 2024-07-08
cli did create web

diehuxx:
Add basic CLI to create and print a did web document

### DIFF
--- a/crates/web5/src/dids/methods/did_web/mod.rs
+++ b/crates/web5/src/dids/methods/did_web/mod.rs
@@ -1,14 +1,17 @@
 mod resolver;
 
 use super::{MethodError, Result};
-use crate::{crypto::jwk::Jwk, dids::{
-    data_model::{document::Document, verification_method::VerificationMethod},
-    did::Did,
-    resolution::{
-        resolution_metadata::{ResolutionMetadata, ResolutionMetadataError},
-        resolution_result::ResolutionResult,
+use crate::{
+    crypto::jwk::Jwk,
+    dids::{
+        data_model::{document::Document, verification_method::VerificationMethod},
+        did::Did,
+        resolution::{
+            resolution_metadata::{ResolutionMetadata, ResolutionMetadataError},
+            resolution_result::ResolutionResult,
+        },
     },
-}};
+};
 use resolver::Resolver;
 
 #[derive(Clone)]
@@ -37,7 +40,7 @@ impl DidWeb {
 
         Ok(DidWeb {
             did: Did::new(&did)?,
-            document
+            document,
         })
     }
 
--- a/crates/web5/src/dids/methods/did_web/mod.rs
+++ b/crates/web5/src/dids/methods/did_web/mod.rs
@@ -13,6 +13,7 @@ use crate::{
     },
 };
 use resolver::Resolver;
+use url::Url;
 
 #[derive(Clone)]
 pub struct DidWeb {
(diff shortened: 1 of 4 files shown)


nitro-neal (approved):
Awesome!

nitro-neal:
if you have the output handy can you put here?

This is amazing!

crates/web5/src/dids/methods/did_web/mod.rs
```diff
@@ -18,6 +21,29 @@ pub struct DidWeb {
 }
 
 impl DidWeb {
+    pub fn new(domain: &str, public_jwk: Jwk) -> Result<Self> {
+        let did = format!("did:web:{}", domain);
+
+        let verification_method = VerificationMethod {
+            id: format!("{}#key-0", did),
+            r#type: "JsonWebKey2020".to_string(),
```
KendallWeihe, line 29: @nitro-neal didn't we standardize this to just `JsonWebKey`?

crates/web5_cli/src/dids/create.rs
```diff
@@ -63,8 +68,23 @@ impl Commands {
...
                 print_portable_did(portable_did, no_indent, json_escape);
             }
-            Commands::Web { domain: _ } => {
-                println!("🚧 not currently supported 🚧");
+            Commands::Web {
+                domain,
+                no_indent,
+                json_escape,
+            } => {
+                let private_jwk = Ed25519Generator::generate();
```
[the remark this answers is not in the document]
KendallWeihe: agreed, on the private key input (portable DID slightly different use case I haven't thought through -- creating a new DID with an existing DID), but yeah not a priority at this moment

crates/web5/src/dids/methods/did_web/mod.rs
```diff
@@ -18,6 +21,29 @@ pub struct DidWeb {
 }
 
 impl DidWeb {
+    pub fn new(domain: &str, public_jwk: Jwk) -> Result<Self> {
+        let did = format!("did:web:{}", domain);
+
+        let verification_method = VerificationMethod {
+            id: format!("{}#key-0", did),
```
KendallWeihe, line 28: I'm slightly troubled by the hard coding of `#key-0` here, but actually moreso with the assumption that all did:web's shall be created initially with a verification method, because the [did:web spec](https://w3c-ccg.github.io/did-method-web/) doesn't assert requirements for the existence of a verification method. This is in contrast with did:dht and did:jwk which both assert requirements of the verification method at time of creation. 

However, we do have the requirement of:

> There must be at least one Verification Method in each DID Document.

(from [our web5-spec](https://github.com/TBD54566975/web5-spec/blob/main/spec/did.md#did-document-data-model)).

So in this case, I think it's right to embrace our spec, because I support the idea that creation of a `did:web` without at least one verification method is non-sensical.

I do wonder if instead we should use a JWK thumbprint for the fragment, but I can see you're using the examples from the did:web spec as inspiration here, which use `#key-0`.

So alright, I've talked myself into this being the ideal way forward. I'm going to update to APID to reflect the constructor here.

KendallWeihe: @diehuxx I have updated the APID, reminder to pull latest.

decentralgabe: I am of two minds here:
1. I think DIDs without verification methods are silly
2. The spec allows for them and so should we

So, we have a few options on how we can handle this
1. Be prescriptive with create but lenient with read/our data model 

This means we can consume DID webs in the wild that may not have VMs, but our impl will only facilitate the creation of DID webs with VMs


2. Allow DID webs to be created / read according to the did:web spec (no VMs needed)

3. Mandate a VM is used to create a DID Web

Here we run into the comments you've raised @KendallWeihe -- what should the first KID be? `#key-0` is a conventional and reasonable choice, and we could expose an option for a user to override it


----------------------------------

zooming out again this really depends on how we're conceptualizing the rust SDK -- is it TBD's opinionated usage of standards? or is it an open source project that provides relatively unopinionated/plain implementations of standards?

KendallWeihe: > Be prescriptive with create but lenient with read/our data model

This seems to be the optimal compromise. We offer support at the data model (and we have an [open PR](https://github.com/TBD54566975/web5-rs/pull/233) on this, which at the moment is serving as a reminder to get to when we have the bandwidth), and then our functions offer an opinionated view. I suppose we can offer an override for the `#key-0` value, but I've been hesitant to offer any optional parameters this until we are further along in our implementation for the sake of agility.

crates/web5_cli/src/dids/create.rs
```diff
@@ -17,7 +17,12 @@ pub enum Commands {
         json_escape: bool,
     },
     Web {
+        #[arg(long)]
```
KendallWeihe, line 20:
```suggestion
```

The domain is a required parameter, so not an `arg` (you can just delete the line). 

Also, thoughts on calling it URL because it could include the HTTP path such as in [Example 5 here](https://w3c-ccg.github.io/did-method-web/#example-creating-the-did-with-optional-path). I'm thinking, name it `url`, and then handle the following use cases:

1. If they pass in `http://some-domain.com` then the `http` would throw an error, b/c needs to be `https`
2. With the caveat, `http` is allowed if the domain name is `localhost` (or `127.0.0.1`)
3. If they pass in `https://some-domain.com` then that's fine, but we have to splice off the `https://`
4. If they pass in the `did.json` file name (as the trailing path part), that's fine but it must be spliced off before creating the DID URI
5. If they pass in the `.well-known` use case in the `url` then it's also spliced off

Basically my thinking is, enable the developer to copy/paste exactly the endpoint they're expecting to host the DID Document at but then ensure the DID URI is spec conformant.

crates/web5/src/dids/methods/did_web/mod.rs
diehuxx: Sounds like the consensus is to leave `#key-0` as the default for new and add the ability to add a custom value in the future. Here's a ticket tracking this: https://github.com/TBD54566975/web5-rs/issues/267

crates/web5_cli/src/dids/create.rs
diehuxx: Added a commit to cover these cases

crates/web5_cli/src/dids/create.rs
```diff
@@ -63,8 +68,60 @@ impl Commands {
...
+            Commands::Web {
+                domain,
+                no_indent,
+                json_escape,
+            } => {
+                let private_jwk = Ed25519Generator::generate();
+                let mut public_jwk = private_jwk.clone();
+                public_jwk.d = None;
+
+                let valid_url = if domain.starts_with("http://") || domain.starts_with("https://") {
```
KendallWeihe, line 80: this is really amazing code @diehuxx 👏 

so good, in fact, I'm going to move it inside the `DidWeb::new()` function, so that we enable this DX not just for the CLI experience but also for application-level use cases. I feel confident exposing this code such that we can expose to developers, "send in any endpoint and we'll create a did:web instance out of it"

also I spy a slight bug in the case wherein both a `.well-known` and `did.json` are passed in, like in [Example 4 here](https://w3c-ccg.github.io/did-method-web/#create-register)

PR #265, opened 2024-07-11
Add support for Ed25519 signing with composite key

KendallWeihe:
Although Ed25519 private keys [are 32 bytes](https://www.rfc-editor.org/rfc/rfc8032#section-5.1.5), I've come to learn some implementations (namely the golang `crypto/ed25519`, but perhaps the JS implementation too?) build private keys as composite keys which contain the public key as well. This PR add supports for cases wherein the key material is a composite key.


KendallWeihe:
I'm not entirely sure I want to merge this. This would enable interop for keys generated in golang, but we don't strictly have a need for that right now, and there's an argument to be made that golang shouldn't be producing Ed25519 JWK's w/ 64 byte `d` values.

PR #268, opened 2024-07-15
Add developer documentation to README

ALRubinger:
Added:

* Environment setup (via Hermit)
* Build commands
* System limitations (Apple Silicon only)

...done as part of inventorying on the way to adding GH Actions CI; should be useful to more devs as they onboard to the project

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,146 @@
 # Work in progress
 
-This repo is not ready for consumption, and is under heavy development
+This repo is not ready for consumption, and is under heavy development. Right now the Rust core bindings work only for Apple Silicon systems.
+
+## Development Prerequisites
+
+### Hermit
+
+This project uses hermit to manage tooling like the Rust compiler, Java Development Kit and Maven project management system.
+See [this page](https://cashapp.github.io/hermit/usage/get-started/) to set up Hermit on your machine - make sure to
+download the open source build and activate it for the project.
+
+Once you've installed Hermit and before running builds on this repo,
+run from the root:
+
+```shell
+source ./bin/activate-hermit
+```
+
+This will set your environment up correctly in the
+terminal emulator you're on. Executing `just` commands should "just work", no
+matter the underlying tooling used (ie. `rustc`, `cargo`, `mvn`, `java`, etc).
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,3 +1,146 @@
 # Work in progress
 
-This repo is not ready for consumption, and is under heavy development
\ No newline at end of file
+This repo is not ready for consumption, and is under heavy development. Right now the Rust core bindings work only for Apple Silicon systems.
```
nitro-neal, line 3: works for 4 architectures 
image

https://github.com/TBD54566975/web5-rs/blob/main/.github/workflows/bound-ci-kt.yml

README.md
```diff
@@ -1,3 +1,146 @@
...
+This project uses hermit to manage tooling like the Rust compiler, Java Development Kit and Maven project management system.
+See [this page](https://cashapp.github.io/hermit/usage/get-started/) to set up Hermit on your machine - make sure to
+download the open source build and activate it for the project.
+
+Once you've installed Hermit and before running builds on this repo,
+run from the root:
+
+```shell
+source ./bin/activate-hermit
+```
```
nitro-neal, line 18: is this the 'default hermit' readme? Isn't it set up so you just cd in a directory and hermit runs this automagically?

nitro-neal (approved):
Thanks! Great first addition to the Readme

README.md
KendallWeihe: That automation occurs via [shell hooks](https://cashapp.github.io/hermit/usage/shell/) -- might be we should make not of it here in the README for dev's who are willing to setup that automation

KendallWeihe (approved):
lfg! this is great!

@ALRubinger we should talk through supporting beyond apple-silicon, but not a top priority at this moment. You're probs aware but right now `tbdex-rs/bound/kt` project is setup to support 4 different target environments, but everything is stuffed in right now which leads to undesirable bandwidth consequences (large jar file, large binaries in git history, build times etc.). Fine for now, but room for improvement. I'm guessing you're already thinking through these matters so just making you aware that I'm aware 😉

PR #269, opened 2024-07-25
Make key_manager a public data member on BearerDid

KendallWeihe:
We need access to the key manager over in tbdex land


nitro-neal (approved):
✅

PR #270, opened 2024-07-25
Add BearerDid.fromPortableDid() and PortableDid.fromJsonString()

KendallWeihe:
Also a bit of cleanup in that we no longer need to recreate the `KeyManager` and `Signer` interfaces since the UniFFI is built with `[WithForeign]`


KendallWeihe:
Moving to a draft, we may not go through with this, at least for the time being to avoid disruption

PR #274, opened 2024-07-29
Add CLI build workflow

KendallWeihe:
- Add GitHub Workflow yml for building the CLI executable for all four targets
- Add the `crates/web5_cli/build` directory full of build scripts for all four targets 
- Add `crates/web5_cli/install.sh` script

That last bit is such that we can simply do this one-liner across (most) Linux and macOS environments

```shell
curl -L https://raw.githubusercontent.com/TBD54566975/web5-rs/crates/web5_cli/install.sh | bash -s v0.0.3
```

### DIFF
--- a/.github/workflows/build-cli.yml
+++ b/.github/workflows/build-cli.yml
@@ -1,0 +1,73 @@
+name: Build CLI
+
+on:
+  push:
+    branches:
+      - main
+      - kendall/cli-build
+  pull_request:
+
+jobs:
+  build_aarch64_apple_darwin:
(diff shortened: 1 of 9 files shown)


KendallWeihe:
This is a WIP branch for building the CLI in the pipeline

For now I'm using it to build the artifacts, download the artifacts and manually trigger a GitHub Release with the binaries attached. Most of this is a temporary measure, with regards to hosting the executable, but the pipeline & build scripts will likely endure.

nitro-neal (approved):
awesome stuff

PR #278, opened 2024-08-07
Add generalized OS/arch support to Justfile

andresuribe87:
This PR enhances the Justfile to add generalized OS/arch support for a more versatile development setup and binding process. It includes the following changes:
- Detection and setup of Rust targets based on current OS and architecture.
- Adjustments to the Justfile to support different environments.

With these changes, the development environment should now properly configure itself for both macOS and Linux environments, targeting both ARM and x86_64 architectures.


frankhinek:
Nice addition.  Looks like still WIP since tests are failing but like the direction.

andresuribe87:
Closing as this was mostly testing goose's capabilities.

PR #279, opened 2024-08-12
Build byte code files in CI, remove from git tree

KendallWeihe:
Previously we have been checking-in the byte code files, but this is undesirable for many reasons. Most of the work here is a rip from https://github.com/TBD54566975/tbdex-rs/pull/96

I've also removed the `jitpack.yml` because that will no longer function with the removal of the byte code files from source control.

@ALRubinger @leordev I went as far as to run Maven tests, but could we go ahead and add snapshot deployments? I don't want to go the full 9-yards and add releases, but if we could go ahead and deploy snapshots then that would be a big win for cross-project integrations (since I'm removing the JitPack integration here).


KendallWeihe:
@leordev @ALRubinger I've added you as CODEOWNERS, lmk if you'd prefer to not be

.github/workflows/ci.yml
```diff
@@ -1,6 +1,18 @@
...
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version of Kotlin binary to publish to TBD Artifactory. For example "1.0.0-SNAPSHOT". If not supplied, will default to "commit-$shortSHA-SNAPSHOT" where $shortSHA is the shortform commit SHA. Must end in "-SNAPSHOT".'
+        required: false
+        default: "0.0.0-SNAPSHOT"
+  workflow_call:
+    inputs:
+      version:
+        description: 'Version of Kotlin binary to publish to TBD Artifactory. For example "1.0.0-SNAPSHOT". If not supplied, will default to "commit-$shortSHA-SNAPSHOT" where $shortSHA is the shortform commit SHA. Must end in "-SNAPSHOT".'
```
leordev, line 13: wrt `Must end in "-SNAPSHOT".` -- I don't see any check to ensure it ends with `-SNAPSHOT` in the `resolve_version` step, but I'm assuming this is fine since `tbdex-rs` is also like that.

leordev (approved):

.github/workflows/ci.yml
KendallWeihe: Good catch. Fixed in latest commit!

KendallWeihe: Also opened this https://github.com/TBD54566975/tbdex-rs/issues/112

.github/workflows/ci.yml
```diff
@@ -10,49 +22,193 @@ env:
   CARGO_TERM_COLOR: always
   # Make sure CI fails on all warnings, including Clippy lints
   RUSTFLAGS: "-Dwarnings"
+  TBDEX_SDK_LOG_LEVEL: debug
```
nitro-neal, line 25: `TBDEX_SDK_LOG_LEVEL` should be called `WEB5_SDK_LOG_LEVEL` here

KendallWeihe: nice catch, fixed!

nitro-neal (approved):

PR #280, opened 2024-08-12
Move sys target initialization to inside UniFFI.kt

KendallWeihe:
This is the wholly remove the possibility of code paths being introduced which fail to first call `SystemTarget.set()`. We ran into this during integration testing of `tbdex-rs/bound/kt` wherein the Kotlin `companion object`'s offered routes for which the call to `SystemTarget.set()` wasn't made, and therefore the loading of the byte code failed in non-macOS aarch64 (the default) settings.

This requires us to augment the code-gen'd `UniFFI.kt` file in our `Justfile`, but all things considered, this is the least invasive (known solution) which is also the most robust.

### DIFF
--- a/bound/kt/src/main/kotlin/web5/sdk/rust/SystemTarget.kt
+++ b/bound/kt/src/main/kotlin/web5/sdk/rust/SystemTarget.kt
@@ -2,59 +2,47 @@ package web5.sdk.rust
 
 import java.io.File
 
-object SystemTarget {
-    @Volatile
-    private var isSet = false
+internal val logLevel = System.getenv("WEB5_SDK_LOG_LEVEL")?.lowercase()
 
-    fun set() {
-        val logLevel = System.getenv("WEB5_SDK_LOG_LEVEL")?.lowercase()
+internal fun log(message: String) {
+    if (logLevel == "debug") {
+        println("web5 sdk SystemArchitecture $message")
+    }
+}
 
-        fun log(message: String) {
-            if (logLevel == "debug") {
-                println("web5 sdk SystemArchitecture $message")
-            }
-        }
+internal fun detectSystemTarget(): String {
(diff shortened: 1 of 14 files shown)


KendallWeihe:
> but to triple check can get this working on the cloud

I don't follow. In [this run](https://github.com/TBD54566975/web5-rs/actions/runs/10354745441/job/28661097706?pr=280) you can see the log in the results, `Successfully loaded shared library for web5_uniffi_x86_64_unknown_linux_gnu` which proves the byte code file for Linux GNU (which is the non-default) successfully loaded, is that what you're referring to?

bound/kt/src/main/kotlin/web5/sdk/rust/SystemTarget.kt
```diff
@@ -2,59 +2,47 @@ package web5.sdk.rust
...
-object SystemTarget {
-    @Volatile
-    private var isSet = false
+internal val logLevel = System.getenv("WEB5_SDK_LOG_LEVEL")?.lowercase()
 
-    fun set() {
-        val logLevel = System.getenv("WEB5_SDK_LOG_LEVEL")?.lowercase()
+internal fun log(message: String) {
+    if (logLevel == "debug") {
+        println("web5 sdk SystemArchitecture $message")
```
nitro-neal, line 9: log vs println

nitro-neal: oh I see this is our loglevel flag,

nitro-neal (approved):
Awesome! Much cleaner

I guess the next level on how to handle this would be to figure out how to write a programatic function in the Uniffi.kt file instead of sed, but sed seems to work find and much less code!

KendallWeihe:
> instead of sed

Yeah, we may be able to find something slightly more robust, but yeah seems like `sed` is a pragmatic solution right now

PR #283, opened 2024-08-14
rfc3339, json, Web5Error, and VerifiableCredential 

KendallWeihe:
- Add `json.rs` module
- Add `rfc3339.rs` module
- Add `Web5Error` in the `errors.rs` module 
- Redo `VerifiableCredential::create()`
- Bind & implement Kotlin code for VC create
- Rename `RustCoreError` to `Web5Errror`
- Write full unit test for VC create for both Rust and Kotlin


Justfile
```diff
@@ -6,7 +6,7 @@ setup:
   source bin/activate-hermit
   git submodule update --init --recursive
   if [[ "$(cargo 2>&1)" == *"rustup could not choose a version of cargo to run"* ]]; then
-    rustup default 1.78.0
+    rustup default 1.80.0
```
KendallWeihe, line 9: we get LazyLock included now! https://blog.rust-lang.org/2024/07/25/Rust-1.80.0.html#lazycell-and-lazylock

crates/web5_cli/src/vcs.rs
```diff
@@ -63,7 +63,7 @@ impl Commands {
                 };
 
                 let now = SystemTime::now();
-                let vc = VerifiableCredential::new(
+                let vc = VerifiableCredential::create(
```
KendallWeihe, line 66: TODO: ensure the CLI isn't broken with these changes

crates/web5/src/rfc3339.rs
```diff
@@ -0,0 +1,55 @@
...
+) -> std::result::Result<SystemTime, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let s = String::deserialize(deserializer)?;
+    let datetime = chrono::DateTime::parse_from_rfc3339(&s).map_err(serde::de::Error::custom)?;
+    Ok(datetime.with_timezone(&Utc).into())
+}
+
+pub(crate) fn serialize_option_system_time<S>(
```
nitro-neal, line 28: in my head `optional` `serialize_option_system_time`  sounds better but I'll leave it up to you

KendallWeihe: I agree! done

crates/web5_cli/src/vcs.rs
KendallWeihe: done

KendallWeihe:
Tests failing, but will fix here shortly

bound/kt/src/test/kotlin/web5/sdk/TestHelper.kt
```diff
@@ -0,0 +1,22 @@
+package web5.sdk
+
+import java.nio.file.Files
+import java.nio.file.Paths
+
+class UnitTestSuite(name: String) {
+    val tests: MutableList<String>
+
+    init {
+        val path = Paths.get("../../tests/unit_test_cases/$name.json")
```
nitro-neal, line 10: This almost needs to be copied in the binary or at least in the bound directory (kind of like schemas) because if people pull it in maven this would not work..

However, people who pull in maven would not execute theses tests so it could be fine? Just some food for thought

bound/kt/src/test/kotlin/web5/sdk/vc/VerifiableCredentialTest.kt
```diff
@@ -0,0 +1,286 @@
...
+            when (e) {
+                is Web5Exception.Exception -> {
+                    assertEquals(e.msg, "parameter error issuer id must not be empty")
+                }
+                else -> {
+                    fail("Caught an unexpected exception type")
+                }
+            }
+        }
+    }
```
nitro-neal, line 131:
```suggestion
@Test
fun testIssuerStringMustNotBeEmpty() {
    testSuite.include()

    val emptyIssuer = Issuer.StringIssuer("")

    val exception = assertFailsWith<Web5Exception.Exception> {
        VerifiableCredential.create(emptyIssuer, CREDENTIAL_SUBJECT, VerifiableCredentialCreateOptions())
    }

    assertEquals("parameter error issuer id must not be empty", exception.msg)
}
```

bound/kt/src/test/kotlin/web5/sdk/vc/VerifiableCredentialTest.kt
```diff
@@ -0,0 +1,286 @@
...
+        val vc = VerifiableCredential.create(ISSUER, CREDENTIAL_SUBJECT, VerifiableCredentialCreateOptions())
+        assertEquals(ISSUER, vc.issuer)
+    }
+
+    @Test
+    fun test_issuer_object_id_must_not_be_empty() {
+        this.testSuite.include()
+        val issuer = Issuer.ObjectIssuer("", "Example Name")
+
+        try {
```
nitro-neal, line 145: if `assertFailsWith` works as the other one can update here too

bound/kt/src/test/kotlin/web5/sdk/vc/VerifiableCredentialTest.kt
```diff
@@ -0,0 +1,286 @@
...
+        assertEquals(issuanceDate, vc.issuanceDate)
+    }
+
+    @Test
+    fun test_issuance_date_must_be_now_if_not_supplied() {
+        this.testSuite.include()
+        val vc = VerifiableCredential.create(ISSUER, CREDENTIAL_SUBJECT, VerifiableCredentialCreateOptions())
+
+        val now = Date()
+        val oneSecondAgo = Date(now.time - 1000)
```
nitro-neal, line 269: maybe set to 1 year or something incase this github action nondeterministically lags at the wrong time or something lol

crates/web5_cli/src/vcs.rs
```diff
@@ -49,7 +49,7 @@ impl Commands {
                     Some(i) => i.to_string(),
                     None => match &portable_did {
                         Some(p) => p.did_uri.to_string(),
-                        None => String::default(),
+                        None => panic!("either --issuer or --portable-did must be supplied"),
```
nitro-neal, line 52: did you mean to keep the panic in?

crates/web5/src/credentials/verifiable_credential_1_1.rs
```diff
@@ -128,85 +118,100 @@ pub struct VerifiableCredential {
...
+            if !types.contains(&BASE_TYPE.to_string()) {
+                types.insert(0, BASE_TYPE.to_string());
+            }
 
-        Self {
-            context: context_with_base,
+            types
+        };
+
+        let id = options
```
nitro-neal, line 191: image
now we get into the philosophical discussion on "what it is to be a vc"

Should we let people put anything they want and create "invalid" vcs

Or should add stuff like this to a validate at the end and "dont let them create" whatever they want

crates/web5/src/credentials/verifiable_credential_1_1.rs
```diff
@@ -128,85 +118,100 @@ pub struct VerifiableCredential {
...
+        Ok(Self {
+            context,
             id,
-            r#type: type_with_base,
+            r#type,
             issuer,
-            issuance_date,
-            expiration_date,
+            issuance_date: options.issuance_date.unwrap_or_else(SystemTime::now),
+            expiration_date: options.expiration_date,
```
nitro-neal, line 201: should we let them create an "invalid" vc that is expired out of the gate.

I think what we have is fine actually, people using this should be aware of what they are doing, 

and once they sign and verify then it will instantly be invalid so they can catch it there but yea things to think about

crates/web5/src/credentials/verifiable_credential_1_1.rs
```diff
@@ -128,85 +118,100 @@ pub struct VerifiableCredential {
...
             credential_subject,
-        }
+        })
     }
 
-    pub fn sign(&self, bearer_did: &BearerDid) -> Result<String> {
+    pub fn sign(&self, bearer_did: &BearerDid) -> ResultOld<String> {
```
nitro-neal, line 206: ResultOld?

nitro-neal (approved):
I'm super onboard with the `CreateOptions` it keeps the apid clean and tight 🥇

bound/kt/src/test/kotlin/web5/sdk/TestHelper.kt
KendallWeihe: fair point, but should be fine for now. as I understand it, our maven artifact does not include tests

bound/kt/src/test/kotlin/web5/sdk/vc/VerifiableCredentialTest.kt
KendallWeihe: very nice! done on all of 'em!

bound/kt/src/test/kotlin/web5/sdk/vc/VerifiableCredentialTest.kt
KendallWeihe: yup done, it's `assertThrows` btw

bound/kt/src/test/kotlin/web5/sdk/vc/VerifiableCredentialTest.kt
KendallWeihe: meh, we'd just re-run the test

maybe I could make it 10 seconds tho, that seems reasonable to me... done!

crates/web5_cli/src/vcs.rs
KendallWeihe: yeah I think it's a slightly better error DX than what the call to `::create()` will return which is just that the issuer string can't be non-empty

remember, this is in the CLI so panic is fine, preferred actually

crates/web5/src/credentials/verifiable_credential_1_1.rs
KendallWeihe: I definitely lean towards being more constrictive than not, but I think we've struck the right amount

[the web5 spec](https://tbd54566975.github.io/web5-spec/#verifiable-credentials-v11-data-model) recommends a format:

> A URI representing a unique identifier for the credential. RECOMMENDED to be of form urn:uuid:3978344f-8596-4c3a-a978-8fcaba3903c5

so we generate in accordance with that format

I'd be down to do more validation, but idk what ruleset we could add here other than "non-empty". `MUST NOT have more than one value` is impossible to validate AFAIK. `MUST be a URI` could be validated, but that seems heavy handed.

it's great food for thought! I welcome any additional ideas!

crates/web5/src/credentials/verifiable_credential_1_1.rs
KendallWeihe: yeah I thought about that but I lean towards enabling them to create an date which is already expired, may be useful is weird edge cases, and the spec doesn't specifically make an assertion except, to your point, vc-jwt verification

crates/web5/src/credentials/verifiable_credential_1_1.rs
KendallWeihe: lol I want to streamline all error cases into the single `Web5Error` instead of the uber-fragmented and all-over-the-place error system we currently have

it's worth noting this is just a rename on the import

KendallWeihe: added this https://github.com/TBD54566975/web5-rs/issues/289

KendallWeihe:
> I'm super onboard with the CreateOptions it keeps the apid clean and tight 🥇

Same, I'm going to iterate through everything and do the same. Feel free to beat me to it @nitro-neal @diehuxx!

Issue #277, opened 2024-08-06
Aggregate open work items


KendallWeihe:
I'm doing the same [over in `tbdex-rs`](https://github.com/TBD54566975/tbdex-rs/issues/108); it's time to put the project management hat on, now that we delivered a feature-complete variant of tbDEX, so that we can strategize upcoming work.

KendallWeihe:
For starters, I would like to spend a day or two reviewing the APID in comparison to `web5-go` and try and converge closer to `web5-go`. We haven't touched this repo in about a month, but IIRC I chose to diverge a bit more than I think is comfortable to most from our old suite of SDKs. For that reason, I think it's well-intentioned and considerate to first and foremost try and bring us closer to the old suite of SDKs; the closer we are to the old SDKs, the easier the transition will be (namely, in updating our existing set of developer guides).

Thereafter...

- Missing VC features: schema, evidence, status list (& maybe more -- do a review in relation to what's in `web5-spec`)
- CI/CD
  - Take heavy inspo from [tbdex-rs](https://github.com/TBD54566975/tbdex-rs) (as, `tbdex-rs` is the most up-to-speed project)
  - Add byte code build jobs (& remove byte code files from git repo)
  - Ensure FOSSA checks pass
  - Ensure tests are running (across both rust & kt)
    - Ensure existing suite of test vectors are integrated
  - Add windows as a build target
  - Publish to Maven central
  - Publish to crates.io
- Debug logging
  - Add debug logging like we did over in tbdex-rs
  - Add version to debug logs
- Consider streamlining all errors into a single `Web5SdkError` like we do at the uniffi layer
  - Rename the `RustCoreError` inside the `web5_uniffi` project to `Web5SdkError`
- Write all `bound/kt` code (a lot is missing)
  - Be sure to follow same patterns used over in `tbdex-rs/bound/kt`
- Extend CLI features
  - Consider interactive DX for some features, such as creating a DID Doc w/ multiple services or VMs
  - The CLI should roughly match the APID, so there are a ton of features we can add here
  - Consider not requiring a Portable DID experience and instead adding integrations into secure solutions
- Write READMEs
- Enable full `secp256k1` support
  - Add the other two missing ([see here](https://did-dht.com/registry/index.html#key-type-index))
- Duplicates from [tbdex-rs 108](https://github.com/TBD54566975/tbdex-rs/issues/108)
[...]

PR #285, opened 2024-08-15
Add debug logging

diehuxx:
Add macro for debug logging and env var to github actions, patterned off tbdex-rs (https://github.com/TBD54566975/tbdex-rs/pull/111/files). Addresses the line of this issue "Add debug logging like we did over in tbdex-rs" https://github.com/TBD54566975/web5-rs/issues/277.

Open Question: In the line "Add version to debug logs" of this issue https://github.com/TBD54566975/web5-rs/issues/277, does that refer to the web5 crate version?


nitro-neal (approved):

KendallWeihe:
> Add version to debug logs 

What I _really_ want is the git commit sha, but that would be require integration into the pipeline building. I would settle for the crate version but that's less good because there are less assurances... that could easily fall through the cracks. Anyways, neither are strictly necessary at this moment, @diehuxx could you create a Issue on the backlog with all of the necessary context for this?

KendallWeihe:
Adding a git commit hash in the debug log is the most bulletproof way to debug production issues at scale. Because then you just open git and you have strong assurances you're reading the source code of the exact artifact that is running.

crates/web5/Cargo.toml
```diff
@@ -15,6 +15,7 @@ josekit = "0.8.6"
 jsonpath-rust = "0.5.1"
 jsonschema = { version = "0.18.0", default-features = false }
 k256 = { version = "0.13.3", features = ["ecdsa", "jwk"] }
+lazy_static = "1.5.0"
```
KendallWeihe, line 18: can we update the Justfile to 1.80.0 and use the new LazyLock? I'm doing that over here https://github.com/TBD54566975/web5-rs/pull/283/files#diff-8b2fda406bc73cf365197448d20bd9ea1d05db17e9b8dca69881ad9dd9755f5d (scroll down to the tests)

crates/web5/src/lib.rs
```diff
@@ -4,3 +4,30 @@ pub mod dids;
...
 #[cfg(test)]
 mod test_helpers;
+
+lazy_static::lazy_static! {
+  pub(crate) static ref LOG_LEVEL: Option<String> = {
+      std::env::var("WEB5_SDK_LOG_LEVEL").ok()
+  };
+}
+
+pub(crate) mod logging {
```
KendallWeihe, line 14: could we create an Issue on the backlog to analyze all the relevant areas where debug logging makes sense and subsequently add it? this is a great starting point but we should also _use_ the things

diehuxx: for sure
https://github.com/TBD54566975/web5-rs/issues/287

crates/web5/build.rs
```diff
@@ -0,0 +1,18 @@
...
+        .expect("Failed to execute git command");
+
+    // Convert the output to a string
+    let git_hash = String::from_utf8(output.stdout).expect("Invalid UTF-8 sequence");
+
+    // Remove the newline character from the commit hash
+    let git_hash_trimmed = git_hash.trim();
+
+    // Pass the commit hash to the compiler as an environment variable
+    println!("cargo:rustc-env=WEB5_GIT_COMMIT_HASH={}", git_hash_trimmed);
```
KendallWeihe, line 17: ayy lfg I wonder if this works all the way through to the kt code just like this

I forgot rust has great build script features

diehuxx: i'd be surprised if it works in KT, so we might need to set up something separate in the KT logger. i don't know as much about KT compilation flags, so that might take me a hot minute

diehuxx: Updated to add the commit hash in kt logs as well

KendallWeihe: so instead, what if in this `build.rs` you added a line of source code inside `lib.rs` that uses the `log_dbg` below to print the git commit hash? that way it's inside the byte code itself such that each sdk we bind to doesn't need to re-implement

crates/web5/src/lib.rs
```diff
@@ -4,3 +4,30 @@ pub mod dids;
 
 #[cfg(test)]
 mod test_helpers;
+
+lazy_static::lazy_static! {
```
KendallWeihe, line 8: could we use the `LazyLock` or `LazyCell` here and then remove the addition of `lazy_static` in the `Cargo.toml`

crates/web5/build.rs
KendallWeihe: come to think of it, this may be a good use case for a procedural macro; write a proc macro, maybe call it `log_git_sha` and then add it to `lib.rs` (`log_git_sha!()`) and it is automatically code-gen's at compile time

crates/web5/src/lib.rs
diehuxx: Upd8d

crates/web5/build.rs
diehuxx: TL;DR it is much simpler to write a separate logging function in Kotlin.

I wasn't able to find a use case for proc macros other than log_dbg. We already have access to the git SHA at compile time via the compile flag `rustc-env=WEB5_GIT_COMMIT_HASH`.

Also, as far as I'm aware, it's not possible to pass a rust macro through UniFFI and call it in the bound language -- in this case, Kotlin. A macro cannot be invoked at runtime, which is necessary for calling it as a bound function in Kotlin. The rust compiler cannot codegen via the macro during Kotlin's compiletime. Also, `log_dbg` is polymorphic, it cannot be described in UDL even if it could be called at runtime.

Another alternative I explored is passing the git SHA through UniFFI. Unfortunately, this approach also came up short. UniFFI does not support passing constants through the language bindings.

This leaves us with two options. First, we can implement a separate logging function in Kotlin. I prefer this. Second, we implement `log_dbg()` as a function, not a macro. Implementing `log_dbg` as a function allows us to pass it across through the language bindings.

KendallWeihe: Like this https://github.com/TBD54566975/web5-rs/pull/290

Lmk whatcha think! I envision more procedural macros in our future so some of this is foundation laying. I think it's slightly preferable because it reduces the overhead for additional bindings and pushes more to the core layer.

PR #284, opened 2024-08-14
Web5 test vectors

nitro-neal:

PR #288, opened 2024-08-16
Consolidate test vectors, add nextest filter

KendallWeihe:
The original driving force for this was [a failing CI](https://github.com/TBD54566975/web5-rs/actions/runs/10409533656/job/28829338878?pr=283). The CI was failing because our new `UnitTestSuite` solution isn't compatible with our `nextest` configuration (best guess is `nextest` is parallelizing things in a way wherein our solution doesn't work; `cargo test` works fine so [our `test` CI job](https://github.com/TBD54566975/web5-rs/actions/runs/10409533656/job/28829339480?pr=283) is passing without issues). We could probably find a way to configure the `nextest` runner/harness such that it is indeed compatible, or we could use an external state file/env-var. But, I think we should consolidate test vectors regardless (IMO we think of test vectors as a singular suite of tests so it makes sense for them to exist in a single module), and if we do, then we can use the `nextest` filter feature to only execute the test vectors in our `rust-test-vectors` CI job.

@nitro-neal lmk what you think (please note, this PR is a merge into #283)


crates/web5/src/test_vectors.rs
```diff
@@ -0,0 +1,148 @@
...
+            #[serde(rename = "didDocument")]
+            did_document: Option<Document>,
+            #[serde(rename = "didDocumentMetadata")]
+            did_document_metadata: DocumentMetadata,
+            #[serde(rename = "didResolutionMetadata")]
+            did_resolution_metadata: ResolutionMetadata,
+        }
+
+        #[test]
+        fn test_did_jwk_resolve() {
```
KendallWeihe, line 56: also slight rename's in the two test names, but everything else is copy/pasta

KendallWeihe:
if we move forward with this, then we need to rebase to retrieve the changes I just merged from https://github.com/TBD54566975/web5-rs/pull/284

KendallWeihe:
> if we move forward with this, then we need to rebase to retrieve the changes I just merged from https://github.com/TBD54566975/web5-rs/pull/284

Alright this is done actually, rebase and moved the latest did:dht test vector into the single module as well

crates/web5/src/test_vectors.rs
```diff
@@ -0,0 +1,210 @@
...
+                assert_eq!(
+                    metadata_error, expected_error,
+                    "Document resolution metadata does not match. Expected '{:?}' but got '{:?}'.",
+                    expected_error, metadata_error
+                );
+            }
+        }
+    }
+
+    mod presentation_definition {
```
nitro-neal, line 162:
```suggestion
    mod presentation_exchange {
```



needs 
image

nitro-neal (approved):
yea thanks! this is much better. I may have to make a change in the test runner depending on how much the xml changes but it should be quick and easy.

Thanks!

crates/web5/src/test_vectors.rs
KendallWeihe: nice catch, fixed

PR #290, opened 2024-08-19
Add web5_proc_macros with git_sha proc macro

KendallWeihe:


KendallWeihe:
I think we should actually push this further and move the entirety of `log_dbg` inside of `web5_proc_macros` https://github.com/TBD54566975/web5-rs/pull/285#discussion_r1721935107 

I'm throwing it back over to @diehuxx for this though

nitro-neal (approved):

PR #292, opened 2024-08-19
add sign and verify test vectors

nitro-neal:
Updating web5-spec and adding ed25519 test vectors


nitro-neal:
this pr goes along with this one - https://github.com/TBD54566975/web5-spec/pull/172

bound/kt/src/main/kotlin/web5/sdk/crypto/verifiers/Ed25519Verifier.kt
```diff
@@ -0,0 +1,27 @@
+package web5.sdk.crypto.verifiers
+
+import web5.sdk.crypto.keys.Jwk
+import web5.sdk.rust.Ed25519Verifier as RustCoreEd25519Verifier
+
+class Ed25519Verifier : Verifier {
+    private val rustCoreVerifier: RustCoreEd25519Verifier
+
+    constructor(privateKey: Jwk) {
+        this.rustCoreVerifier = RustCoreEd25519Verifier(privateKey)
```
KendallWeihe, line 10: I think with the latest on `main` you'll need to rebase and then this'll be...

```suggestion
        this.rustCoreVerifier = RustCoreEd25519Verifier(privateKey.rustCoreJwkData)
```

bound/kt/src/test/kotlin/web5/sdk/TestVectorsTest.kt
```diff
@@ -0,0 +1,104 @@
...
+)
+
+class Web5TestVectorsCryptoEd25519 {
+
+    data class SignTestInput(
+        val data: String,
+        val key: TestVectorJwk,
+    )
+
+    data class TestVectorJwk(
```
KendallWeihe, line 30: I know we're not testing the JWK in this test, but ideally we could use the real `Jwk`, but to your point on the other PR, we may need to add #297 first?

KendallWeihe (approved):
well done! I added one comment with respect to rebasing onto `main` which needs addressed but after that this is good to go!

crates/web5/src/test_vectors.rs
```diff
@@ -207,4 +208,100 @@ mod test_vectors {
...
+                let data = hex_string_to_byte_array(input.data.to_string());
+                let result = signer.sign(&data);
+
+                if matches!(vector.errors, Some(true)) {
+                    assert!(result.is_err(), "Expected an error, but signing succeeded");
+                } else {
+                    let signature = result.expect("Signing should not fail");
+
+                    // Convert the signature to a hex string
+                    let signature_hex = byte_array_to_hex_string(&signature);
```
diehuxx, line 252: I'm surprised this works. ed25519 signatures are different every time because they use a random nonce to prevent computing the private key if the same private key is used for multiple signatures

crates/web5/src/test_vectors.rs
```diff
@@ -207,4 +208,100 @@ mod test_vectors {
...
+                let expected_output = vector.output;
+
+                let verifier = Ed25519Verifier::new(input.key);
+
+                let data = hex_string_to_byte_array(input.data.to_string());
+                let signature = hex_string_to_byte_array(input.signature.to_string());
+
+                let result = verifier.verify(&data, &signature);
+
+                let is_valid = result.expect("Verification should not fail");
```
diehuxx, line 280: what's the difference between Err() and Ok(false) for `verifier.verify`?

bound/kt/src/test/kotlin/web5/sdk/TestVectorsTest.kt
nitro-neal: yup after 297 is in we can use it

crates/web5/src/test_vectors.rs
nitro-neal: chad says:
Ok(false): The operation itself succeeded, but the verification result is false. This means that the signature is valid in terms of format and process, but it does not match the data or is otherwise invalid according to the verification logic.

and yea then err is an exception basically if I'm understnading rust correctly

nitro-neal: oh and as far as the implementation yea just verified false is invalid signature vs catastrophic failure

crates/web5/src/test_vectors.rs
nitro-neal: oh no.. 
This works in every sdk

crates/web5/src/test_vectors.rs
KendallWeihe: FYI in #300 I'm moving in favor of the monad pattern so `verify` returns a `void` and utilizes the error case for cryptographic failure 

no actionable items in this PR, just making you aware (and I'll be sure to remediate test vectors prior to merging #300 )

crates/web5/src/test_vectors.rs
nitro-neal: I think we are good - https://cryptobook.nakov.com/digital-signatures/eddsa-and-ed25519

decentralgabe: see [Section 4.2.5 on Deterministic Nonce Generation](https://eprint.iacr.org/2020/823.pdf)

> Ed25519 uses deterministic signing which removes the need for fresh random numbers during the signing process. This does not lead to any particular consequences for our security analysis since we model the key derivation function as a random oracle. However, it is well known not to reduce security

PR #294, opened 2024-08-20
Add feature complete Jwk

KendallWeihe:
- Add comprehensive unit test coverage for JWK's compute thumbprint
- Add compute thumbprint uniffi binding
- Add `ToInner*` and `ToOuter*` for the kt key manager and signer (disintermediates the uniffi code gen)
- Add new `Web5Error:: DataMember` error variant -- signifies a semantic error for a given instances data members (ex. an empty string is a valid `String` type in Rust but semantically invalid)

### DIFF
--- a/bound/kt/src/main/kotlin/web5/sdk/crypto/keys/Jwk.kt
+++ b/bound/kt/src/main/kotlin/web5/sdk/crypto/keys/Jwk.kt
@@ -1,10 +1,44 @@
 package web5.sdk.crypto.keys
 
+import web5.sdk.rust.Jwk as RustCoreJwk
 import web5.sdk.rust.JwkData as RustCoreJwkData
 
 /**
  * Partial representation of a [JSON Web Key as per RFC7517](https://tools.ietf.org/html/rfc7517).
  * Note that this is a subset of the spec.
  */
+data class Jwk (
+    val alg: String? = null,
+    val kty: String,
+    val crv: String,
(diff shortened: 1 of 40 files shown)


bound/kt/src/main/kotlin/web5/sdk/crypto/keys/Jwk.kt
```diff
@@ -1,10 +1,44 @@
 package web5.sdk.crypto.keys
 
+import web5.sdk.rust.Jwk as RustCoreJwk
 import web5.sdk.rust.JwkData as RustCoreJwkData
 
 /**
  * Partial representation of a [JSON Web Key as per RFC7517](https://tools.ietf.org/html/rfc7517).
  * Note that this is a subset of the spec.
  */
+data class Jwk (
```
nitro-neal, line 10: by the way, why don't we have a `kid` field? we have them in the key for the test vectors - https://github.com/TBD54566975/web5-spec/blob/main/test-vectors/crypto_ed25519/sign.json

nitro-neal (approved):

bound/kt/src/main/kotlin/web5/sdk/crypto/keys/Jwk.kt
KendallWeihe: Because we haven't have an immediate application use case for it

Created #297 feel free to add it!

PR #299, opened 2024-08-20
Support X25519 in did:dht verification method

diehuxx:


KendallWeihe:
nice! I was actually just working on #300 but there may be a good amount of conflicts with this, so I'll pause

@diehuxx two things:

1. generator, signer, and verifier for x25519 would be 💯 
2. full unit test coverage like I have over in #300 

I kinda lean towards lets go ahead and get this merged into main so that I can continue on with #300, and then after #300 is merged then you can follow up with the above two items?

diehuxx:
> 1. generator, signer, and verifier for x25519 would be 💯

What's the use case for generator, signer, and verifier? It looks like that's not in our public API in the API Design doc, so I was keeping support scoped to just did:dht. This way web5-rs did:dht implementation can accept DIDs with x25519 but we don't need to maintain a generator, signer, and verifier.

KendallWeihe:
> What's the use case for generator, signer, and verifier? 

We've committed to supporting the four cryptographic primitives under web5 as defined in the spec here https://tbd54566975.github.io/web5-spec/#cryptographic-primitives

diehuxx:
@KendallWeihe Makes sense. I'll add a commit updating the API Design doc and can add the tests and addl utils in a follow up PR

KendallWeihe:
> in a follow up PR

Okay excellent. At this rate, it'll probably be tomorrow before I get #300 in, but if you want to get a head start, you can safely go ahead and branch off of that branch (once I rebase after your merge here)

KendallWeihe (approved):
👏

diehuxx:
@KendallWeihe Looks like X25519 is [not used for digital signatures](https://www.cryptosys.net/pki/manpki/pki_eccsafecurves.html#:~:text=The%20NIST%2FSEC%20elliptic%20curves%20can%20be%20used%20for%20both,cannot%20be%20used%20for%20signatures.), so it may not have notions of signing and verification that correspond with the existing `Signer` and `Verifier` traits. What purposes do you want X25519 to support?

decentralgabe:
the main function of x25519 is to generate a shared secret...this is surfaced through [scalar multiplication functions](https://www.rfc-editor.org/rfc/rfc7748.html#section-5). see the go std lib for an example: https://pkg.go.dev/golang.org/x/crypto/curve25519

and code example

```go
package main

import (
	"crypto/rand"
	"encoding/hex"
	"fmt"
	"golang.org/x/crypto/curve25519"
)

func main() {
	// Generate Alice's private key
	alicePrivate := make([]byte, curve25519.ScalarSize)
	_, err := rand.Read(alicePrivate)
	if err != nil {
		panic(err)
	}

	// Generate Alice's public key
	var alicePublic [curve25519.PointSize]byte
	curve25519.ScalarBaseMult(&alicePublic, (*[32]byte)(alicePrivate))

	// Generate Bob's private key
	bobPrivate := make([]byte, curve25519.ScalarSize)
	_, err = rand.Read(bobPrivate)
	if err != nil {
		panic(err)
	}

	// Generate Bob's public key
	var bobPublic [curve25519.PointSize]byte
	curve25519.ScalarBaseMult(&bobPublic, (*[32]byte)(bobPrivate))

	// Alice computes shared secret
	var aliceShared [curve25519.PointSize]byte
	curve25519.ScalarMult(&aliceShared, (*[32]byte)(alicePrivate), &bobPublic)

	// Bob computes shared secret
	var bobShared [curve25519.PointSize]byte
	curve25519.ScalarMult(&bobShared, (*[32]byte)(bobPrivate), &alicePublic)

	// Print the shared secrets
	fmt.Printf("Alice's shared secret: %s\n", hex.EncodeToString(aliceShared[:]))
	fmt.Printf("Bob's shared secret:   %s\n", hex.EncodeToString(bobShared[:]))
}
```

I don't know if we're doing anything related to shared secret generation or key agreement in our stack today...so it may be fine to generate an X25519 key but not do anything with it yet

KendallWeihe:
Word, nice! so sounds like we're just going with `X25519Generator` and that's enough

PR #300, opened 2024-08-20
Use error for Verifier::verify(), Add ed25519 unit tests, Add Web5Error::Crypto

KendallWeihe:
Part of the iterative process of bringing the entirety of the code base up to speed, this PR focuses on the DSA components (signer, verifier, & ed25519).

- Introduce `Web5Error::Crypto` variant (& also `::Unknown` and `::Encoding` variants)
- Change `Verifier::verify()` function signature to utilize error for failure case (as opposed to boolean)
- Add comprehensive unit test coverage for Ed25519
- Write the Kotlin code for these matters, plus unit test coverage

### DIFF
--- a/bound/kt/src/main/kotlin/web5/sdk/crypto/verifiers/Verifier.kt
+++ b/bound/kt/src/main/kotlin/web5/sdk/crypto/verifiers/Verifier.kt
@@ -1,5 +1,30 @@
 package web5.sdk.crypto.verifiers
 
 import web5.sdk.rust.Verifier as RustCoreVerifier
+import web5.sdk.rust.Web5Exception
 
-typealias Verifier = RustCoreVerifier
+/**
+ * Set of functionality required to implement to be a compatible DSA verifier.
+ */
+interface Verifier {
+    /**
+     * Execute the verification of the signature against the payload by using the encapsulated public key material.
+     *
+     * @param message the data which was signed over.
+     * @param signature the signature over the message.
+     * @throws Web5Exception in the case of a failed verification.
+     */
+    fun verify(message: ByteArray, signature: ByteArray)
+}
+
+internal class ToOuterVerifier(private val rustCoreVerifier: RustCoreVerifier) : Verifier {
+    override fun verify(message: ByteArray, signature: ByteArray) {
+        rustCoreVerifier.verify(message, signature)
+    }
+}
+
+internal class ToInnerVerifier(private val verifier: Verifier) : RustCoreVerifier {
+    override fun verify(message: ByteArray, signature: ByteArray) {
+        verifier.verify(message, signature)
+    }
+}
--- a/bound/kt/src/test/kotlin/web5/sdk/Web5TestVectorsTest.kt
+++ b/bound/kt/src/test/kotlin/web5/sdk/Web5TestVectorsTest.kt
@@ -21,6 +21,7 @@ import web5.sdk.crypto.verifiers.Ed25519Verifier
 import web5.sdk.dids.Document
 import web5.sdk.rust.DocumentMetadataData
 import web5.sdk.rust.ResolutionMetadataData
+import web5.sdk.rust.Web5Exception
 
 class Web5TestVectorsTest {
 
@@ -110,20 +111,21 @@ class Web5TestVectorsTest {
                 val ed25519Jwk = Jwk(
                     kty = testVectorJwk.kty,
                     crv = testVectorJwk.crv,
-                    d = null,
+                    d = testVectorJwk.d,
                     x = testVectorJwk.x,
                     alg = null,
                     y = null
                 )
                 val verifier = Ed25519Verifier(ed25519Jwk)
 
-                if (vector.errors == true) {
-                    assertThrows(Exception::class.java) {
+                if (vector.errors == true || vector.output == false) {
+                    assertThrows(Web5Exception.Exception::class.java) {
                         verifier.verify(inputByteArray, signatureByteArray)
                     }
                 } else {
-                    val verified = verifier.verify(inputByteArray, signatureByteArray)
-                    assertEquals(vector.output, verified)
+                    assertDoesNotThrow {
+                        verifier.verify(inputByteArray, signatureByteArray)
+                    }
                 }
             }
         }
--- a/bound/kt/src/test/kotlin/web5/sdk/crypto/verifiers/Ed25519VerifierTest.kt
+++ b/bound/kt/src/test/kotlin/web5/sdk/crypto/verifiers/Ed25519VerifierTest.kt
@@ -1,41 +1,127 @@
 package web5.sdk.crypto.verifiers
 
-import org.junit.jupiter.api.Test
-import org.junit.jupiter.api.Assertions.assertTrue
-import org.junit.jupiter.api.Assertions.assertFalse
+import org.junit.jupiter.api.*
+import org.junit.jupiter.api.Assertions.*
+import org.junit.jupiter.api.fail
+import web5.sdk.UnitTestSuite
+import web5.sdk.crypto.Ed25519Generator
 import web5.sdk.crypto.keys.Jwk
 import web5.sdk.crypto.signers.Ed25519Signer
-import web5.sdk.rust.ed25519GeneratorGenerate as rustCoreEd25519GeneratorGenerate
+import web5.sdk.rust.Web5Exception
 
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
 class Ed25519VerifierTest {
 
-    @Test
-    fun `test verifier with valid signature`() {
-        val privateJwk = rustCoreEd25519GeneratorGenerate()
-        val ed25519Signer = Ed25519Signer(Jwk.fromRustCoreJwkData(privateJwk))
+    private val testSuite = UnitTestSuite("ed25519_verify")
 
-        val message = "abc".toByteArray()
-        val signature = ed25519Signer.sign(message)
+    @AfterAll
+    fun verifyAllTestsIncluded() {
+        if (testSuite.tests.isNotEmpty()) {
+            println("The following tests were not included or executed:")
+            testSuite.tests.forEach { println(it) }
+            fail("Not all tests were executed! ${this.testSuite.tests}")
+        }
+    }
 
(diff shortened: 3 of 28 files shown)


KendallWeihe:
Converting to Draft -- @diehuxx will overlap with your #299 so we'll have to coordinate

KendallWeihe:
Main items remaining are to write the kt code, but also this will have conflicts with #299 so I need to coordinate remediation

bound/kt/src/test/kotlin/web5/sdk/Web5TestVectorsTest.kt
```diff
@@ -21,6 +21,7 @@ import web5.sdk.crypto.verifiers.Ed25519Verifier
 import web5.sdk.dids.Document
```
KendallWeihe, line 21: fyi @nitro-neal changes here

crates/web5/src/test_vectors.rs
```diff
@@ -214,6 +214,7 @@ mod test_vectors {
         use crate::crypto::dsa::ed25519::{Ed25519Signer, Ed25519Verifier};
```
KendallWeihe, line 214: also fyi @nitro-neal here too

bound/kt/src/main/kotlin/web5/sdk/crypto/verifiers/Verifier.kt
```diff
@@ -1,5 +1,30 @@
...
+    override fun verify(message: ByteArray, signature: ByteArray) {
+        rustCoreVerifier.verify(message, signature)
+    }
+}
+
+internal class ToInnerVerifier(private val verifier: Verifier) : RustCoreVerifier {
+    override fun verify(message: ByteArray, signature: ByteArray) {
+        verifier.verify(message, signature)
+    }
+}
```
diehuxx, line 30: Having trouble understanding the intended use for these based on the name. `ToYadayada` sounds like a function name. If the intention is to convert from `Verifier` to `RustCoreVerifier`, why use a class instead of a function?

bound/kt/src/test/kotlin/web5/sdk/crypto/verifiers/Ed25519VerifierTest.kt
```diff
@@ -1,41 +1,127 @@
...
+
+        assertTrue(verifyResult.isSuccess, "Verification should succeed with a valid signature")
+    }
+
+    @Test
+    fun test_with_private_key() {
+        this.testSuite.include()
+
+        val (_, privateJwk) = generateKeys()
+        val verifier = Ed25519Verifier(privateJwk) // this is not allowed
```
diehuxx, line 53: If this is not allowed, should this line throw?

crates/web5/src/credentials/verifiable_credential_1_1.rs
```diff
@@ -515,18 +515,9 @@ impl JosekitJwsVerifier for JoseVerifier {
...
-            .map_err(|e| JosekitError::InvalidSignature(e.into()))?;
-
-        match result {
-            true => Ok(()),
-            false => Err(JosekitError::InvalidSignature(
-                // 🚧 improve error message semantics
-                DsaError::VerificationFailure("ed25519 verification failed".to_string()).into(),
-            )),
-        }
+            .map_err(|e| JosekitError::InvalidSignature(e.into()))
```
diehuxx, line 520: nice, much simpler.

crates/web5/src/dids/methods/did_dht/mod.rs
```diff
@@ -22,7 +22,7 @@ use crate::{
 };
 use std::sync::Arc;
 
-pub mod bep44;
+pub(crate) mod bep44;
```
diehuxx, line 25: why not just omit `pub` entirely? doesn't look like we're using this anywhere else in the crate

crates/web5/src/errors.rs
```diff
@@ -10,6 +13,16 @@ pub enum Web5Error {
...
+    #[error("cryptography error {0}")]
+    Crypto(String),
+    #[error("encoding error {0}")]
+    Encoding(String),
+}
+
+impl From<DecodeError> for Web5Error {
+    fn from(err: DecodeError) -> Self {
+        Web5Error::Encoding(err.to_string())
+    }
```
diehuxx, line 25: I remember we chatted about this at standup but I had to think it through again. Writing the train of thought here in case anyone else has the same question:

why manually implement `From<>` for error conversions instead of using `#[from]` from `thiserror`? Because `#[from]` puts the the source error as the associated payload -- e.g. `Web5Error::Encoding(base64::DecodeError)`. Whereas we want to use a `String` as the associated payload because this displays better in error stack traces in bound languages.

crates/web5/src/crypto/dsa/ed25519.rs
```diff
@@ -109,29 +117,347 @@ impl Ed25519Verifier {
...
+    use super::*;
+    use crate::{test_helpers::UnitTestSuite, test_name};
+    use general_purpose::URL_SAFE_NO_PAD;
+    use std::sync::LazyLock;
+
+    mod generate {
+        use super::*;
+
+        static TEST_SUITE: LazyLock<UnitTestSuite> =
+            LazyLock::new(|| UnitTestSuite::new("ed25519_generate"));
```
diehuxx, line 176: Are we committing to using rustc 1.80.0 features in our tests for this crate?

bound/kt/src/main/kotlin/web5/sdk/crypto/verifiers/Verifier.kt
KendallWeihe: Yup it's a great question! The overarching requirement is to enable developers to build-their-own `Verifier`, in this case, but it also applies to other interfaces such as `KeyManager` or `Signer`. Basically, these two concrete (and `internal` -- that's important) implementations are translation layers to-and-from the UniFFI generated interfaces. We don't want the consuming apps to be directly utilizing the UniFFI code gen, so we have to wrap around it, and to wrap an interface we have to concretize it with a simple wrapper implementation.

crates/web5/src/dids/methods/did_dht/mod.rs
KendallWeihe: there are some public items inside the `bep44` crate which aren't in the APID. I guess we could add bep44 features to our APID, but I lean towards not doing so, for now at least

crates/web5/src/crypto/dsa/ed25519.rs
KendallWeihe: no def not, I'm just behind on your MSRV work

bound/kt/src/test/kotlin/web5/sdk/crypto/verifiers/Ed25519VerifierTest.kt
KendallWeihe: I had the same thought, but we have a test vector ([here](https://github.com/TBD54566975/web5-spec/blob/7c4e7fc5206ee3f50c05383e61be0046623f7dc9/test-vectors/crypto_ed25519/verify.json#L110)) which asserts this behavior, so if we did it in the constructor then we'd be doing it twice. 

@nitro-neal any thoughts on changing the test vector?

crates/web5/src/errors.rs
KendallWeihe: The matter we discussed was actually about the `serde_json::Error` which we were forced to implement the `From` trait for because it doesn't derive the `Clone`, but it's a good question because it's brought me to the _real_ reason I think we should go the `From<>` route instead of the `#[from]` route, which is encapsulation/coupling. With `#[error(transparent)]Encoding(#[from] DecodeError)` we do indeed have a richer set of information and stronger types, however, we leak `base64` abstractions across our publicly accessible API. Or said differently, we're strongly coupling our crate to `base64` as well as the consumers of our crate. But `base64` is an encapsulated detail, and if it's not and we choose to change to a different dependency in the future then we would be introducing a breaking change. So like, I would be in favor of the `#[from]` route for errors which are strictly encapsulated, but anything which is exposed to the publicly accessible API should be our own types. 

So basically, TL;DR prevent abstraction bleed.

crates/web5/src/credentials/verifiable_credential_1_1.rs
KendallWeihe: yeah I think so too, I'm sold on the monad pattern

crates/web5/src/dids/methods/did_dht/mod.rs
diehuxx: I'd prefer not to add bep44 to the APID. I was thinking we could do `mod bep44` instead of `pub(crate) bep44` since I don't think we need bep44 outside of did_dht, unless I'm missing something

crates/web5/src/crypto/dsa/secp256k1.rs
```diff
@@ -3,9 +3,12 @@ use crate::errors::Result;
 use crate::errors::Web5Error;
 use base64::{engine::general_purpose, Engine as _};
 
+#[cfg(test)]
 pub struct Secp256k1Generator;
 
+#[cfg(test)]
 impl Secp256k1Generator {
+    #[cfg(test)]
```
diehuxx, line 11: If this is meant to be part of the public API, why only build it for testing?

crates/web5/src/dids/methods/did_dht/mod.rs
KendallWeihe: oh right, I agree! done

crates/web5/src/crypto/dsa/secp256k1.rs
KendallWeihe: yup, done!

KendallWeihe: fyi @diehuxx in [the commit here](https://github.com/TBD54566975/web5-rs/pull/300/commits/89a7bd71a8ef93678677ccb3c750c3878be92533) I had to make the secp256k1 mod fully public because CI is setup to fail on warnings https://github.com/TBD54566975/web5-rs/actions/runs/10497204019/job/29079498494

just a heads up to that latest change

diehuxx (approved):

PR #301, opened 2024-08-21
Generate X25519 keys

diehuxx:
Follow up from https://github.com/TBD54566975/web5-rs/pull/299

### DIFF
--- a/crates/web5/Cargo.toml
+++ b/crates/web5/Cargo.toml
@@ -26,7 +26,7 @@ simple-dns = "0.7.0"
 thiserror = { workspace = true }
 url = "2.5.0"
 uuid = { workspace = true }
-x25519-dalek = { version = "2.0.1", features = ["getrandom"] }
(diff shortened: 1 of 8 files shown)


KendallWeihe:
I guess since X25519 isn't a signature alg then it shouldn't be under the DSA namespace, yeah?

KendallWeihe:
we should remove `X25519Signer` and `X25519Verifier` from the APID too yeah?

diehuxx:
Makes sense to me. Any preference on namespace?

KendallWeihe:
> Any preference on namespace

It sounds like this is about key exchange, right? So idk maybe `key_exchange`? or else maybe just in the root crypto

decentralgabe:
[key exchange](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/x25519/) is common terminology

KendallWeihe (approved):

PR #303, opened 2024-08-21
Set MSRV to 1.74.0

KendallWeihe:


diehuxx (commented):
Looks like the Justfile isn't being edited in this PR so several parts of the CI are still using 1.80.0

crates/web5/rust-toolchain.toml
diehuxx, line 1: we don't need this file if we're using the same rustc version throughout the repo

KendallWeihe: I think we should probably keep it as cargo takes this file as the highest priority, so in other words it's the source of truth. It may prevent accidental situations where a developer sets their local toolchain to something other than what's in this file. Also, it's probably good practice to enable developers to come to this repo with their own rust setup and not have to rely on the Justfile.

KendallWeihe:
> Looks like the Justfile isn't being edited in this PR so several parts of the CI are still using 1.80.0

I set the Justfile to 1.74.0 https://github.com/TBD54566975/web5-rs/pull/303/files#diff-2f90408c2b0302b1cdb7f5d634114750837f940fa82d13057d9c18d0170a7e5c

diehuxx (approved):

PR #307, opened 2024-08-23
Add DidWeb::create() and DidWeb::resolve(), add unit tests, bind to kt

KendallWeihe:


bound/kt/src/test/kotlin/web5/sdk/dids/methods/web/DidWebTests.kt
```diff
@@ -1,50 +1,276 @@
...
+
+            // temporarily removing @context from document, need to rework Document type
+            bearerDid.document.context = null
+
+            mockWebServer.enqueue(
+                MockResponse()
+                    .setResponseCode(200)
+                    .addHeader("Content-Type", "application/json")
+                    .setBody(Json.stringify(bearerDid.document))
+            )
```
diehuxx, line 265: We could stricten this test by only returning did doc from `/.well-known/did.json `. If you wanna go bananas, we could add a test case for when did uri specifies the path

docs/API_DESIGN.md
```diff
@@ -624,27 +623,24 @@ CLASS DidJwkCreateOptions
...
 >
-> The `CONSTRUCTOR(domain: string, public_jwk: Jwk)` does not publish the DID Document to a host, but merely creates the instance of the `did:web` in the local scope. 
+> The `create` does not publish the DID Document to a host, but creates the instance of the `did:web` `BearerDid` in the local scope. 
 
 ```pseudocode!
 CLASS DidWeb
-  CONSTRUCTOR(domain: string, public_jwk: Jwk)
-  CONSTRUCTOR(uri: string)
+  STATIC METHOD create(domain: string, options: DidWebCreateOptions?): BearerDid
   STATIC METHOD resolve(uri: string): ResolutionResult
```
diehuxx, line 631: outside the scope of this PR, but curious if you've given thought to whether did web resolve, did dht resolve, and did dht publish should be async?

diehuxx (approved):

nitro-neal (approved):

docs/API_DESIGN.md
KendallWeihe: yeah I think that's right actually, we should do that ... async programming in the binded world maybe needs a bit more thought 🤔 

https://github.com/TBD54566975/web5-rs/issues/309

bound/kt/src/test/kotlin/web5/sdk/dids/methods/web/DidWebTests.kt
KendallWeihe: yeah, I agree with both of those items! but for a later time 😅

PR #308, opened 2024-08-23
Add DidDht create, publish and resolve, add unit tests

KendallWeihe:


bound/kt/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
```diff
@@ -1,78 +1,84 @@
...
+import web5.sdk.dids.BearerDid
 import web5.sdk.dids.ResolutionResult
+import web5.sdk.rust.ServiceData
+import web5.sdk.rust.VerificationMethodData
 import web5.sdk.rust.didDhtResolve as rustCoreDidDhtResolve
-import web5.sdk.rust.DidDht as RustCoreDidDht
-import web5.sdk.rust.Signer as RustCoreSigner
+
+data class DidDhtCreateOptions(
+    val publish: Boolean? = false,
```
nitro-neal, line 12: so we don't publish by default? Should maybe talk to gabe, I'm 95% sure we will run into user error on this a lot. people will do default did dht create and wonder why it is not working for resolve

nitro-neal (approved):
looks great!

bound/kt/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
diehuxx: > people will do default did dht create and wonder why it is not working for resolve

I would argue it is good for people to notice it is not working as they might expect and wonder why, causing them to learn that did:dht is a published did. If they do not understand that aspect of did:dht, they cannot make an informed choice of which did method to use.

KendallWeihe: it actually is set to publish by default [as you can see here](https://github.com/TBD54566975/web5-rs/blob/21a38dba3fef3dbfc1c2596a0200342cc735208b/crates/web5/src/dids/methods/did_dht/mod.rs#L106) 

I was following [the web5-js design](https://github.com/TBD54566975/web5-js/blob/34590b2d50dbfff4c60747ce23c33fdd155a4366/packages/dids/src/methods/did-dht.ts#L616)

but perhaps here I should reinforce that experience by seeing `= true`? what do we think? Rust doesn't have the equivalent syntax so I was following Rust

KendallWeihe: alright, that's changed in the latest commit. I think we should set it as `true` at that layer, even though the default is `true` inside the rust core code (as I linked to above), because developers often look at the function signature to determine default behavior. good catch!

decentralgabe: I support publish by default

KendallWeihe: oh I just realized the way I had it before, there are code paths where the default value you be false actually, even tho I'm using the default value of true inside the rust core code. jumping across languages is hard! thanks for the spot @nitro-neal

decentralgabe (commented):
are these test vectors implemented? https://did-dht.com/#test-vectors

KendallWeihe:
> are these test vectors implemented? https://did-dht.com/#test-vectors

@nitro-neal is following behind with test vectors

it's an interesting question tho, because those specific test vectors are distinct from our web5-spec test vectors, but maybe they shouldn't be?

decentralgabe:
@KendallWeihe yes ... there has been discussion about this for a while 😄 [^1] [^2] [^3] [^4] [^5] [^6] [^7]

[^1]: https://github.com/TBD54566975/web5-go/issues/142
[^2]: https://github.com/TBD54566975/web5-kt/issues/303
[^3]: https://github.com/TBD54566975/web5-js/issues/497
[^4]: https://github.com/TBD54566975/web5-kt/pull/334#discussion_r1714610864
[^5]: https://github.com/TBD54566975/web5-rs/issues/105#issuecomment-2073648662
[^6]: https://github.com/TBD54566975/web5-spec/issues/167
[^7]: https://github.com/TBD54566975/web5-spec/issues/11#issuecomment-1821932704

KendallWeihe:
@decentralgabe brought the receipts! 😆 

We'll get 'er done here https://github.com/TBD54566975/web5-rs/issues/310

PR #311, opened 2024-08-26
Add full Bearer DID implementation, add full unit test coverage

KendallWeihe:


bindings/web5_uniffi_wrapper/src/dids/bearer_did.rs
```diff
@@ -21,10 +24,13 @@ pub struct BearerDidData {
 pub struct BearerDid(pub InnerBearerDid);
 
 impl BearerDid {
-    pub fn new(uri: &str, key_manager: Arc<dyn KeyManager>) -> Result<Self> {
+    pub fn new(did: Did, document: Document, key_manager: Arc<dyn KeyManager>) -> Self {
```
nitro-neal, line 27: eh? we can't do did.document?

nitro-neal: oh right nm, this is the query / params / uri part

nitro-neal: then what benefit do we have for this vs just the string uri? Sometimes the string is more convenient to the user I would guess

KendallWeihe: I agree actually. This is an area where I diverged from the old SDKs so I was moving back in line with them. But I think there is room for both: instantiation from the constituent parts, but also from a DID URI (which encapsulates resolution). If you agree, stub in a ticket on the backlog!

crates/web5/src/dids/bearer_did.rs
```diff
@@ -67,24 +40,183 @@ impl BearerDid {
...
+            std::thread::sleep(std::time::Duration::from_millis(100));
+
+            TEST_SUITE.assert_coverage()
+        }
+
+        #[test]
+        fn test_query_requirements_not_satisfied() {
+            TEST_SUITE.include(test_name!());
+
+            let bearer_did = DidJwk::create(None).unwrap();
```
nitro-neal, line 142: thoughts on create_default()?

```
    pub fn create(options: Option<Options>) -> Self {
        // Your logic here using `options`
    }

    pub fn create_default() -> Self {
        Self::create(None)
    }
    ```

nitro-neal: and then in kotlin we can overload actually have create() and create(options)

KendallWeihe: I agree but would prefer to keep the APID surface area as minimal as possible for the time being

This is a good case study of, if we weren't targeting many languages all with the same API surface, then we could leverage an idiom unique to rust and implement the `Default` trait for the `DidJwk` struct, but I digress

KendallWeihe: in other words, the wiser strategy is to accept a bit more verbosity in the name of supporting many languages, and fine tune it over time

PR #312, opened 2024-08-26
Add full ResolutionResult coverage

KendallWeihe:


crates/web5/src/dids/resolution/resolution_result.rs
```diff
@@ -14,30 +18,30 @@ pub struct ResolutionResult {
...
 }
 
+#[derive(Default)]
+pub struct ResolutionResultResolveOptions {
+    pub did_dht_gateway_url: Option<String>,
+}
+
 impl ResolutionResult {
-    pub fn new(uri: &str) -> Self {
+    pub fn resolve(uri: &str, options: Option<ResolutionResultResolveOptions>) -> Self {
```
nitro-neal, line 27: In a future update we will update so that we dont have this: ResolutionResultResolveOptions

and it will be a confi object inside the did dht resolver?

nitro-neal (approved):

crates/web5/src/dids/resolution/resolution_result.rs
KendallWeihe: that's right, I'm going to follow up right away with that -- choosing not to do it here because it'll be quite a bit of changes which'll bloat this

PR #313, opened 2024-08-26
Add full VC unit test coverage and converge on APID

KendallWeihe:


KendallWeihe:
Moving to _Draft_, requesting feedback on APID proposal from @nitro-neal & @diehuxx

docs/API_DESIGN.md
```diff
@@ -90,11 +92,29 @@ CLASS VerifiableCredential
   PUBLIC DATA credentialSubject: CredentialSubject
 
   CONSTRUCTOR create(issuer: Issuer, credential_subject: CredentialSubject, options: CreateOptions)
+  CONSTRUCTOR decode(vc_jwt: string)
```
nitro-neal, line 95: I remember we went though this before with the decode vs verify vs both on the js side. 

Basically definitely don't have both as it is confusing to the end user. 

I would do the superset of functionally, I would have `verify` and `verifyOptions`.

 But now is always the question:
 
 of if it's expired or something and your verify just throws or returns a bool, how would the user know what is in there if they decided to decode it? 
 
 Could implement a verify results object, but this also causes confusion

nitro-neal: btw in your current impl I think you wanted VerifiableCredentialDecodeOptions on the decode too right?

nitro-neal: I'm reading this wrong, this is a new constructor not VerifiableCredential.decode() ok let me rethink lol

nitro-neal: another option is to have `create_from_vc_jwt(vc_jwt, createFromVcJwtOptions)`

KendallWeihe: > btw in your current impl I think you wanted VerifiableCredentialDecodeOptions on the decode too right?

yes! fixed 👍

KendallWeihe: > Basically definitely don't have both as it is confusing to the end user

This was my driving principal as I was thinking it through as well. However, I will add [that `web5-go` does actually offer both](https://github.com/TBD54566975/web5-go/blob/main/vc/vcjwt.go), so I'm not strictly opposed to it.

If we only have one entrypoint, then it must be named either `verify()` or `decode()`, right? I went with `decode()` with the option `verify: bool` parameter because both features are expected to decode the VC-JWT, but only 1 or the 2 will execute cryptographic verification (& also, @nitro-neal would you bucket semantic data model validation under that same `verify` feature or should we distinguish between validation & verification? So that would be two options `verify: bool` and `validate: bool`?).

KendallWeihe:
@nitro-neal @diehuxx Another question, how do we tie-in the `issuer` field? Right now, I have it implemented such that the `issuer` MUST be set during the `create` call. However, one potential code path in the `sign()` function uses the `BearerDid` instance, so it's possible the developer could set the `issuer` to value A during the call to `create` but then call `sign()` with a different instance of a `BearerDid`.

The solution I'll put forth is defined in this commit I just pushed up https://github.com/TBD54566975/web5-rs/pull/313/commits/b7f8f0c90e4ac7fa56eed1f8a707a4b817358304

docs/API_DESIGN.md
nitro-neal: yea! we went down that road too lol validate vs verify.

I think we should just have verify and it should validate under the hood.

nitro-neal: Actually I think this approach is slightly different because decode is a constructor where as verify could be a method on the object:

```
  CONSTRUCTOR create(issuer: Issuer, credential_subject: CredentialSubject, options: CreateOptions)
  CONSTRUCTOR decode(vc_jwt: string, options: VerifiableCredentialDecodeOptions) // you almost dont need options here, for more clarity this could be called from_vc_jwt or something too.., and then in kotlin or something we could pug sugar around it where we have 2 constructors, one for issuer, cred, and options, another for vcjwt
  METHOD sign(options: VerifiableCredentialSignOptions)
  METHOD verify(options: VerifiableCredentialVerifyOptions) // this validates too
  ```

**Comment by nitro-neal** [comment]
it wont be a valid vc if the issuer is A and the signer is B right?

could just throw if you ever see that state?

File: docs/API_DESIGN.md (hunk 1 above)
Comment (KendallWeihe) [reply docs/API_DESIGN.md]: Okay I like that, and yeah I would propose we go with `from_vc_jwt()` to be consistent with our `from_json_string()`. Then that would raise the question, should we name `sign()` to `to_vc_jwt()` 🤔 

The reason I don't _love_ the design you lay out, is there is a code path from `create()` to `verify()` which is non-sensical because `verify()` is only relevant if the originating information was a VC-JWT. Or, put more concretely, that `verify()` has a signature to verify against. Any thoughts on this? Maybe we just handle that with an error, or maybe if the VC-JWT isn't known to the instance then it only performs data model validation and simply ignores cryptographic verification?

**Comment by KendallWeihe** [comment]
> it wont be a valid vc if the issuer is A and the signer is B right?
> 
> could just throw if you ever see that state?

Yup, agreed, that's what the language in that commit I linked to was implying. Alright cool, that matter is settled 👍

File: docs/API_DESIGN.md [hunk 2]
```diff
@@ -90,11 +92,29 @@ CLASS VerifiableCredential
...
+```pseudocode!
+CLASS VerifiableCredentialDecodeOptions
+  PUBLIC DATA verify: bool?
+```
+
+##### `VerifiableCredentialSignOptions`
+
+> [!NOTE]
+>
+> If a `signer` is specified, then it takes first priority. Else, the `bearer_did` will be used. If a `verification_method_id` is not specified then the first Verification Method in the `bearer_did` will be utilized. Given the `signer` IS NOT specified and the `bearer_did` IS specified, then the `issuer` ID defined in the call to `create()` MUST be equal to value of the `bearer_did`'s DID URI.
```
diehuxx, line 111: this is a little hard to follow. might be better to have multiple sign methods that take different arguments

diehuxx: It looks like the two ways to sign are
1. `sign(signer: Signer)`
2. `sign_with_did(bearer_did: BearerDid, verification_method_id: Option<String>)`

docs/API_DESIGN.md
```diff
@@ -90,11 +92,29 @@ CLASS VerifiableCredential
...
-  METHOD sign(bearer_did: BearerDid): string
-  METHOD sign_with_signer(key_id: string, signer: Signer): string
+  METHOD sign(options: VerifiableCredentialSignOptions)
+```
+
+##### `VerifiableCredentialDecodeOptions`
+
+```pseudocode!
+CLASS VerifiableCredentialDecodeOptions
+  PUBLIC DATA verify: bool?
```
diehuxx, line 104: I usually prefer a modular approach of separate decode() and verify() methods. If we have to combine it, then it seems clearer to use a `verify: bool` instead of `bool?`. If we have to use `bool?`, what is the behavior of `verify == null`? Does it verify?

If we have only one option, why not have it as a param, rather than defining a struct that wraps a boolean?

diehuxx: Is it only possible to verify open calling `decode()`? Or do we plan to also add a separate `verify()` method later, to allow someone to verify after decoding?

docs/API_DESIGN.md
KendallWeihe: I'm cool with breaking it up, however with the caveat of this instead

`sign_with_did(bearer_did: BearerDid, options: VerifiableCredentialSignWithDidOptions)`

The reason being, across all of our SDKs, there are a variety of opinions as to how we should _query_ for the given signer, so we can't commit to only supporting `verification_method_id`.

KendallWeihe:
Closing in favor of #317

PR #317, opened 2024-08-27
Add full VC unit test coverage and converge on APID

KendallWeihe:
Also change `BearerDid`'s `get_signer()` method to just take a `verification_method_id`, for the same reasons we converged on `sign_with_did()` taking a `verification_method_id` -- this is our go to solution for key selection at this time.


docs/API_DESIGN.md
```diff
@@ -90,11 +93,10 @@ CLASS VerifiableCredential
   PUBLIC DATA credentialSubject: CredentialSubject
 
   CONSTRUCTOR create(issuer: Issuer, credential_subject: CredentialSubject, options: CreateOptions)
+  CONSTRUCTOR from_vc_jwt(vc_jwt: string, verify: bool)
```
KendallWeihe, line 96: @nitro-neal @diehuxx I forgot about this yesterday during our sync... do we want to enable developers to bring-their-own `Verifier`? If so how would that fit into this function signature?

KendallWeihe: One more item @nitro-neal @diehuxx, regardless of the above comment, should the default behavior be for cryptographic verification to occur? So with the code as it is currently that would mean `verify: bool = true` (though if that were the case I think we should stick with `bool` values always defaulting to false so we would change the name to `disable_verification: bool`).

diehuxx: We chatted offline, so I'm summarizing our discussion here to document.

IMO, BYO verifier is extraneous. If someone wants to use their own verifier, they can use it after decoding. Passing a custom verifier into the decoder is added complexity in our API and implementation.

crates/web5/src/credentials/verifiable_credential_1_1.rs
diehuxx, line 1: haven't started reading this yet, but 1600 line file should maybe be broken into 2 or 3 files. Even though it's mostly tests it's still 600 lines of app code, which is a lot on its own. we can break it up in a follow up PR if you prefer.

crates/web5/src/credentials/verifiable_credential_1_1.rs
```diff
@@ -205,93 +216,61 @@ impl VerifiableCredential {
...
-        let vc_jwt = josekit::jwt::encode_with_signer(&payload, &header, &jose_signer)?;
-
-        Ok(vc_jwt)
-    }
-
-    pub fn verify(vc_jwt: &str) -> ResultOld<Self> {
-        // this function currently only supports Ed25519
-        let header = josekit::jwt::decode_header(vc_jwt)?;
+    // this function currently only supports Ed25519
+    pub fn from_vc_jwt(vc_jwt: &str, verify: bool) -> Result<Self> {
```
diehuxx, line 220: I'm looking through this function to find ways to break it up into smaller pieces. One candidate is the `if verify` block. There's definitely one or two more that I'll find as I read through

crates/web5/src/credentials/verifiable_credential_1_1.rs
```diff
@@ -205,93 +216,61 @@ impl VerifiableCredential {
...
+                };
+
+                let (jwt_payload, _) = josekit::jwt::decode_with_verifier(vc_jwt, jose_verifier)
+                    .map_err(|e| {
+                        Web5Error::Crypto(format!("vc-jwt failed cryptographic verification {}", e))
+                    })?;
+
+                jwt_payload
+            }
+            false => {
```
diehuxx, line 262: if statement instead of matching against boolean?

crates/web5/src/credentials/verifiable_credential_1_1.rs
```diff
@@ -538,6 +584,33 @@ impl core::fmt::Debug for JoseVerifier {
     }
 }
 
+#[derive(Clone)]
+struct JoseVerifierAlwaysTrue {
```
diehuxx, line 588: What is this for?

diehuxx: As in, why have a verifier that always returns true?

crates/web5/src/credentials/verifiable_credential_1_1.rs
```diff
@@ -383,57 +359,127 @@ impl VerifiableCredential {
...
 
-        validate_vc_data_model(&vc)?;
+        if verify {
+            validate_vc_data_model(&vc)?;
+        }
 
         Ok(vc)
     }
+
+    fn sign_with_signer(&self, key_id: &str, signer: Arc<dyn Signer>) -> Result<String> {
```
diehuxx, line 369: interesting name. not pub though? are we not allowing a generic signer after all?

diehuxx: also, why do we need key_id for this one?

crates/web5/src/credentials/verifiable_credential_1_1.rs
```diff
@@ -316,51 +295,48 @@ impl VerifiableCredential {
...
+                        })?
                         .as_secs()
-                        != exp.duration_since(UNIX_EPOCH)?.as_secs()
+                        != exp
+                            .duration_since(UNIX_EPOCH)
+                            .map_err(|e| {
+                                Web5Error::Unknown(format!("unknown system time error {}", e))
+                            })?
+                            .as_secs()
                     {
```
diehuxx, line 336: this 11 line conditional is a little tough on the eyes. it runs its own conditionals and has two possible return statements all within the conditional part of the `if`

crates/web5/src/credentials/verifiable_credential_1_1.rs
KendallWeihe: if you want to do this in a follow up then feel free! I don't have strong opinions and err on the side of keeping everything together

crates/web5/src/credentials/verifiable_credential_1_1.rs
KendallWeihe: same as my prior comment, I err on keeping everything together as breaking it up is only boilerplate if it only has one use

crates/web5/src/credentials/verifiable_credential_1_1.rs
KendallWeihe: done 👍

crates/web5/src/credentials/verifiable_credential_1_1.rs
KendallWeihe: because josekit constraints

josekit offers a `decode_unsecured` function, but it fails if the jose header includes an `alg`, which ours will in this case. so the alternative would be to decode the header, then mutate it, then create a new jwt token, then pass it into `decode_unsecured`. this just seemed easier. josekit is great, but it does offer a sort of esoteric API surface. 

if you have alternatives then I'm game!

crates/web5/src/credentials/verifiable_credential_1_1.rs
KendallWeihe: due to some reasonable doubts, we're moving forward with solely "sign with did" on the public api surface at this time. we can extend later by adding lower level support for DSA signers and JOSE signers, but for now they aren't strictly necessary.

a `Signer` is a DSA signer, which is distinct from a (proposed) `JoseSigner`. DSA is lower level, it's just about producing a signature for a given input, whereas JOSE is higher level which includes additional semantic data, in our case, a `kid`.

crates/web5/src/credentials/verifiable_credential_1_1.rs
KendallWeihe: simplified in latest commit. lmk whatcha think!

diehuxx:
@KendallWeihe https://github.com/TBD54566975/web5-rs/pull/321

KendallWeihe:
> @KendallWeihe #321. I think this could go further, but I'm opening it now so you can see what I'm thinking

nice! lmk when it's ready for a review

nitro-neal (approved):

PR #318, opened 2024-08-27
Decouple kt data structs from UniFFI codegen

KendallWeihe:
This way we aren't tightly coupled with what UniFFI codegens

Also added `from_json_string` and `to_json_string` to the DID Document

### DIFF
--- a/bound/kt/src/main/kotlin/web5/sdk/crypto/Dsa.kt
+++ b/bound/kt/src/main/kotlin/web5/sdk/crypto/Dsa.kt
@@ -1,0 +1,6 @@
+package web5.sdk.crypto
+
+enum class Dsa {
+    ED25519,
+    SECP256K1;
+}
--- a/bound/kt/src/main/kotlin/web5/sdk/dids/Document.kt
+++ b/bound/kt/src/main/kotlin/web5/sdk/dids/Document.kt
@@ -1,8 +1,106 @@
 package web5.sdk.dids
 
-import web5.sdk.rust.DocumentData as RustCoreDocumentData
+import web5.sdk.crypto.keys.Jwk
 
 /**
  * Representation of a [DID Document](https://github.com/TBD54566975/web5-spec/blob/main/spec/did.md).
  */
-typealias Document = RustCoreDocumentData
+data class Document(
+    val id: String,
+    val context: List<String>?,
+    val controller: List<String>?,
+    val alsoKnownAs: List<String>?,
+    val verificationMethod: List<VerificationMethod>,
+    val authentication: List<String>?,
+    val assertionMethod: List<String>?,
+    val keyAgreement: List<String>?,
+    val capabilityInvocation: List<String>?,
+    val capabilityDelegation: List<String>?,
+    val service: List<Service>?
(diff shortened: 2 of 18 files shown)


bound/kt/src/main/kotlin/web5/sdk/crypto/Dsa.kt
```diff
@@ -0,0 +1,6 @@
+package web5.sdk.crypto
+
+enum class Dsa {
```
nitro-neal, line 3: i see so we have to keep this in sync manually instead of using the rust ones.

I wonder if there can be a dynamic way to do this, create new enum at runtime that matches the names.. idk, but this is fine

bound/kt/src/main/kotlin/web5/sdk/dids/Document.kt
```diff
@@ -1,8 +1,106 @@
...
+import web5.sdk.crypto.keys.Jwk
 
 /**
  * Representation of a [DID Document](https://github.com/TBD54566975/web5-spec/blob/main/spec/did.md).
  */
-typealias Document = RustCoreDocumentData
\ No newline at end of file
+data class Document(
+    val id: String,
+    val context: List<String>?,
```
nitro-neal, line 10: yup the thing we lose is like if we are missing a field or we have something optional and its not optional here or something.

Like Document in theory could get out of sync with RustCoreDocumentData. 

But thats what test vectors and unit tests are for

bound/kt/src/main/kotlin/web5/sdk/dids/ResolutionResult.kt
```diff
@@ -1,30 +1,86 @@
...
+                documentMetadata.canonicalId
+            )
         }
     }
+}
+
+data class ResolutionMetadata(
+    val error: ResolutionMetadataError?
+) {
+    companion object {
```
nitro-neal, line 57: perfect

nitro-neal (approved):

diehuxx:
@KendallWeihe Can you add some discussion about _why_ we do not want to be coupled to uniffi codegen'd data structs? The obvious benefit of using codegen'd structs is less work for us. The benefits of decoupling are not immediately clear.

bound/kt/src/main/kotlin/web5/sdk/crypto/Dsa.kt
diehuxx: > I wonder if there can be a dynamic way to do this

If we find ourselves writing code to codegen based on codegen'd code, we are lost in the sauce.

KendallWeihe:
> add some discussion about _why_ we do not want to be coupled to uniffi codegen'd data structs

The primary reason being I'm not convinced we'll stick with UniFFI, so decoupling enables us to move to something else without introducing breaking changes. We may also move from UniFFI UDL to proc macros, and if we do then it'll almost certainly augment the shape of the codegen. UniFFI is an implementation detail, not the solution itself. Rewriting the data structs & enums may seem like boilerplate but it's negligible effort (an hour or two) for a lot of reward.

PR #319, opened 2024-08-28
WIP Flutter rust bridge

diehuxx:
Current state: when running the flutter app as a macos app, there is a linking error related to reqwest not correctly enabling the system configurations library for Apple devices. Same issue as this https://github.com/fzyzcjy/flutter_rust_bridge/issues/2104, but those solutions aren't quite working.


diehuxx:
closing while we figure out how we want to handle openSSL issue and http in general

PR #320, opened 2024-08-28
cred status list impl

nitro-neal:


crates/web5/src/credentials/status_list_credential.rs
```diff
@@ -0,0 +1,560 @@
...
+
+        let verifiable_credential =
+            VerifiableCredential::create(issuer, credential_subject, Some(vc_options))?;
+
+        Ok(Self {
+            base: verifiable_credential,
+        })
+    }
+
+    /// Checks if a given credential is disabled according to this Status List Credential.
```
diehuxx, line 78: 👍 docs 🙏

crates/web5/src/credentials/status_list_credential.rs
```diff
@@ -0,0 +1,560 @@
...
+    /// Retrieves the value of a specific bit from a compressed base64 URL-encoded bitstring
+    /// by decoding and decompressing a bitstring, then extracting a bit's value by its index.
+    ///
+    /// # Arguments
+    /// * `compressed_bitstring` - A base64 URL-encoded string representing the compressed bitstring.
+    /// * `bit_index` - The zero-based index of the bit to retrieve from the decompressed bitstream.
+    ///
+    /// # Returns
+    /// `true` if the bit at the specified index is 1, `false` if it is 0.
+    fn get_bit(compressed_bitstring: &str, bit_index: usize) -> Result<bool> {
```
diehuxx, line 252: This looks like it's not StatusListCredential-specific, no? Might be better to put this in a separate helper file and move associated tests as well

crates/web5/src/credentials/mod.rs
```diff
@@ -8,6 +8,7 @@ mod josekit;
 mod jwt_payload_vc;
 pub mod presentation_definition;
 mod sign;
+pub mod status_list_credential;
```
diehuxx, line 11: slight preference for
```suggestion
mod status_list_credential;
pub use status_list_credential::{StatusListCredential, STATUS_LIST_CREDENTIAL_CONTEXT, STATUS_LIST_CREDENTIAL_TYPE, STATUS_LIST_2021, STATUS_LIST_2021_ENTRY};
```

crates/web5/src/credentials/status_list_credential.rs
```diff
@@ -0,0 +1,560 @@
...
+        // Check if the status type matches
+        if status.r#type != STATUS_LIST_2021_ENTRY {
+            return Err(Web5Error::Parameter(format!(
+                "unsupported status type: {}",
+                status.r#type
+            )));
+        }
+
+        // Check if the status purpose matches
+        let status_purpose = self
```
diehuxx, line 108: for both `status_purpose` and `encoded_list` we've got these similar, super vertical blocks of code that do "get this string attribute from additional_properties". I think it'd be more readable and reusable to put logic in a private function that doesn't use as many `and_then`s.

crates/web5/src/credentials/status_list_credential.rs
nitro-neal: I think in reality it will only be used for status list credential, its getting the bit from a compressed gziped base64 string, which I think nothing else in the universe uses 😆

crates/web5/src/credentials/status_list_credential.rs
nitro-neal: yup you are right, let me do this

crates/web5/src/credentials/status_list_credential.rs
```diff
@@ -0,0 +1,560 @@
...
+    }
+
+    /// Generates a compressed, base64-encoded bitstring from a list of status list indexes.
+    ///
+    /// # Arguments
+    /// * `status_list_indexes` - A vector of indexes to set in the bitstring.
+    ///
+    /// # Returns
+    /// A `Result` containing the compressed, base64-encoded bitstring, or an error if an index is out of range.
+    fn bitstring_generation(status_list_indexes: Vec<usize>) -> Result<String> {
```
diehuxx, line 210: nit: don't need to take ownership. might avoid unnecessary cloning
```suggestion
    fn bitstring_generation(status_list_indexes: &[usize]) -> Result<String> {
```

crates/web5/src/credentials/status_list_credential.rs
```diff
@@ -0,0 +1,560 @@
...
+            properties.insert(
+                "type".to_string(),
+                JsonValue::String(STATUS_LIST_2021.to_string()),
+            );
+            properties.insert(
+                "encodedList".to_string(),
+                JsonValue::String(base64_bitstring.clone()),
+            );
+            JsonObject { properties }
+        };
```
diehuxx, line 52: Easier to see that we're just initting a hashmap with a few values. You can also use `HashMap::from()` if you prefer, instead of `into_iter().collect()`.
```suggestion
        let additional_properties = JsonObject {
            properties: [
                ("statusPurpose".to_string(), JsonValue::String(status_purpose)),
                ("type".to_string(), JsonValue::String(STATUS_LIST_2021.to_string())),
                ("encodedList".to_string(), JsonValue::String(base64_bitstring)),
            ].into_iter().collect(),
        };
```

crates/web5/src/credentials/status_list_credential.rs
```diff
@@ -0,0 +1,560 @@
...
+
+        let vc_options = VerifiableCredentialCreateOptions {
+            id: Some(format!("urn:uuid:{}", uuid::Uuid::new_v4())),
+            context: Some(vec![STATUS_LIST_CREDENTIAL_CONTEXT.to_string()]),
+            r#type: Some(vec![STATUS_LIST_CREDENTIAL_TYPE.to_string()]),
+            issuance_date: None,
+            expiration_date: None,
+            credential_status: None,
+            credential_schema: None,
+            evidence: None,
```
diehuxx, line 67:
```suggestion
            ..Default::default()
```

diehuxx (approved):
looks good. all my comments are non blocking

crates/web5/src/credentials/status_list_credential.rs
diehuxx: since it doesn't mention `StatusListCredential`, still seems worth putting in a separate module. Could be in a module within `mod status_list_credential` if you like. Putting it in the `impl StatusListCredential` disrupts the focus.

crates/web5/src/credentials/status_list_credential.rs
nitro-neal: genius.

crates/web5/src/credentials/status_list_credential.rs
```diff
@@ -0,0 +1,533 @@
...
+            .collect(),
+        };
+
+        let credential_subject = CredentialSubject {
+            id: format!("urn:uuid:{}", uuid::Uuid::new_v4()),
+            additional_properties: Some(additional_properties),
+        };
+
+        let vc_options = VerifiableCredentialCreateOptions {
+            id: Some(format!("urn:uuid:{}", uuid::Uuid::new_v4())),
```
KendallWeihe, line 61: you don't necessarily have to pass this in

bound/kt/src/main/kotlin/web5/sdk/vc/StatusListCredential.kt
```diff
@@ -0,0 +1,27 @@
+package web5.sdk.vc
+
+import web5.sdk.Json
+import web5.sdk.rust.StatusListCredential as RustCoreStatusListCredential
+
+data class StatusListCredential(
```
KendallWeihe, line 6: we want this to match the APID so we need to add a `val base VerifiableCredential` here

PR #321, opened 2024-08-28
Decompose verifiable_credential_1_1.rs

diehuxx:
verifiable_credential_1_1.rs is 1600 lines long, so we should break it into smaller focused files that are easier to understand. Here is one way to break it up.


bound/kt/src/test/kotlin/web5/sdk/vc/VerifiableCredentialTest.kt
```diff
@@ -572,6 +572,22 @@ class VerifiableCredentialTest {
 
             assertEquals("claim mismatch: expiration_date", exception.msg)
         }
+    }
+
+    @Nested
+    @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+    inner class DataModelValidation {
```
KendallWeihe, line 579: we want to keep the unit test suites 1:1 to the APID features, because we're testing the public api surface. if we want to add a publicly facing feature for data model validation then we can do this, otherwise it must be embedded in the `from_vc_jwt` like it was

bindings/web5_uniffi/src/web5.udl
```diff
@@ -264,7 +264,7 @@ interface VerifiableCredential {
   constructor(
     string json_serialized_issuer,
     string json_serialized_credential_subject,
-    VerifiableCredentialCreateOptionsData? options
+    VerifiableCredentialCreateOptionsData options
```
KendallWeihe, line 267: this should be optional, the APID is incorrect in its current state but the code was right. unless you think it should be required?

crates/web5/src/credentials/mod.rs
```diff
@@ -1,6 +1,17 @@
...
+mod jwt_payload_vc;
 pub mod presentation_definition;
+mod sign;
 pub mod verifiable_credential_1_1;
 
+pub use credential_subject::CredentialSubject;
+pub use issuer::{Issuer, ObjectIssuer};
+
 #[derive(thiserror::Error, Debug, Clone, PartialEq)]
 pub enum CredentialError {
```
KendallWeihe, line 16: let's do this, since it's not part of the public APID surface, it's an implementation detail which rolls into `Web5Error` (I want to add `Web5Error` to the APID)

```suggestion
pub(crate) enum CredentialError {
```

crates/web5/src/credentials/jose_kit.rs
```diff
@@ -0,0 +1,103 @@
+use std::{fmt::Formatter, sync::Arc};
```
KendallWeihe, line 1: `josekit` is [all one word](https://crates.io/crates/josekit)

diehuxx: fixed

PR #329, opened 2024-08-29
Add windows target support to kt

KendallWeihe:
Running pipeline will have a `kotlin-acceptance-testing` job which runs the full unit test suite on a `windows-latest` os

Closes #189

PR #330, opened 2024-08-29
Build CLI binaries and attach them to release

finn-tbd:
This adds GitHub Action jobs that build cli binaries when a release is created and attaches them to the release. You can see what this looks like on the [latest test release in my branch](https://github.com/finn-tbd/web5-rs/releases/tag/v0.1.4). To use it, simply create a release and the automation will take care of the rest.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -22,7 +22,7 @@ jobs:
           files: web5_cli-aarch64-apple-darwin
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
-  build_x86_64_apple_darwin:
+  build_cli_x86_64_apple_darwin:
     runs-on: macos-12
     name: x86_64-apple-darwin
     steps:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,72 @@
+name: Build web5_cli release artifacts
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  build_cli_aarch64-apple-darwin:
+    runs-on: macos-latest
+    name: aarch64-apple-darwin
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install Rust
+        run: rustup toolchain install stable
+      - name: cargo build
+        run: |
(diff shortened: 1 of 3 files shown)


KendallWeihe:
just as a heads up, I just added a windows target build in #329 -- no need to make it part of this PR, but we should expect to add CLI support for Windows (but that may be multifaceted so can do later if it's excessive here)

.github/workflows/release.yaml
```diff
@@ -0,0 +1,72 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install Rust
+        run: rustup toolchain install stable
+      - name: cargo build
+        run: |
+          cargo build --target aarch64-apple-darwin --release --package web5_cli
+          cp -v target/aarch64-apple-darwin/release/web5_cli web5_cli-aarch64-apple-darwin
+      - name: upload release asset
+        uses: AButler/upload-release-assets@v3.0
```
KendallWeihe, line 20: was thinking about using [the GitHub native action](https://github.com/actions/upload-release-asset?tab=readme-ov-file) for this, but looks like it's no longer maintained. however they do recommend this one https://github.com/softprops/action-gh-release. any thoughts on this vs the AButler one you have here?

KendallWeihe:
@finn-tbd any thoughts on adding the `install.sh` script to the release assets? because then users can `curl` the releases URL for the install script

KendallWeihe (approved):
added a few comments for discussion points, but this lgtm ✅ 

after we merge, could you try to utilize it in this repo? we're not strict on versioning right now, looks like 0.0.5 would be the next version

.github/workflows/release.yaml
finn-tbd: I didnt see a way to use the AButler one to update an existing release. It wanted to be the one to create the release.

finn-tbd:
IMO we should not add `install.sh` to the release assets. We could post the `curl .. | bash` line in the release nodes ("to install or update, run...") but it seems kind of odd to post the same file over and over and if it changed. Also, `install.sh` will hopefully become `brew install ...` eventually

KendallWeihe:
@finn-tbd nice, yeah agreed!

PR #332, opened 2024-08-30
Add VC Credential Schema

KendallWeihe:


bindings/web5_uniffi/src/web5.udl
```diff
@@ -257,10 +257,23 @@ dictionary VerifiableCredentialCreateOptionsData {
   sequence<string>? type;
   timestamp? issuance_date;
   timestamp? expiration_date;
+  CredentialSchemaData? credential_schema;
   string? json_serialized_evidence;
 };
 
 interface VerifiableCredential {
+  [Throws=Web5Error]
+  constructor(
```
diehuxx, line 266: what's this second constructor for?

crates/web5/src/credentials/credential_schema.rs
```diff
@@ -0,0 +1,72 @@
+use super::verifiable_credential_1_1::VerifiableCredential;
+use crate::errors::{Result, Web5Error};
+use jsonschema::{Draft, JSONSchema};
+use reqwest::blocking::get;
```
diehuxx, line 4: another spot for us to think about making async in the future

crates/web5/src/credentials/mod.rs
```diff
@@ -1,4 +1,5 @@
 mod create;
+pub mod credential_schema;
```
diehuxx, line 2: `use credentials::credential_schema::CredentialSchema` vs `use credentials::CredentialSchema`. up to you, i prefer the latter.

diehuxx (approved):

bindings/web5_uniffi/src/web5.udl
KendallWeihe: good callout, this is unnecessary actually. thanks!

crates/web5/src/credentials/mod.rs
KendallWeihe: nice yeah totally agree

nitro-neal (approved):

PR #334, opened 2024-09-03
Decouple from UniFFI code-gen'd Web5Exception

KendallWeihe:
this is the last piece to completely decouple us from uniffi codegen at the api surface 

also fixed the Dsa which was still a lingering coupling

### DIFF
--- a/bound/kt/src/main/kotlin/web5/sdk/crypto/Dsa.kt
+++ b/bound/kt/src/main/kotlin/web5/sdk/crypto/Dsa.kt
@@ -1,6 +1,22 @@
 package web5.sdk.crypto
 
+import web5.sdk.rust.Dsa as RustCoreDsa
+
 enum class Dsa {
     ED25519,
     SECP256K1;
+}
+
+internal fun dsaFromRustCore(rustCore: RustCoreDsa): Dsa {
+    return when (rustCore) {
+        RustCoreDsa.ED25519 -> Dsa.ED25519
+        RustCoreDsa.SECP256K1 -> Dsa.SECP256K1
+    }
+}
+
+internal fun dsaToRustCore(dsa: Dsa): RustCoreDsa {
+    return when(dsa) {
+        Dsa.ED25519 -> RustCoreDsa.ED25519
+        Dsa.SECP256K1 -> RustCoreDsa.SECP256K1
+    }
 }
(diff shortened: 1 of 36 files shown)


diehuxx (approved):

bound/kt/src/main/kotlin/web5/sdk/crypto/Dsa.kt
```diff
@@ -1,6 +1,22 @@
 package web5.sdk.crypto
 
+import web5.sdk.rust.Dsa as RustCoreDsa
+
 enum class Dsa {
     ED25519,
     SECP256K1;
+}
+
+internal fun dsaFromRustCore(rustCore: RustCoreDsa): Dsa {
```
nitro-neal, line 10: perfect

bound/kt/src/main/kotlin/web5/sdk/crypto/Ed25519Generator.kt
```diff
@@ -14,8 +16,14 @@ class Ed25519Generator {
          * @return Jwk the JWK with private key material included.
          */
         fun generate(): Jwk {
```
nitro-neal, line 18: chad says:
You can simplify the generate() function by removing the redundant catch block for the generic Exception, as it will automatically be propagated if not caught.

bound/kt/src/main/kotlin/web5/sdk/crypto/Ed25519Generator.kt
```diff
@@ -14,8 +16,14 @@ class Ed25519Generator {
...
-            val rustCoreJwkData = ed25519GeneratorGenerate()
-            return Jwk.fromRustCoreJwkData(rustCoreJwkData)
+            try {
+                val rustCoreJwkData = ed25519GeneratorGenerate()
+                return Jwk.fromRustCoreJwkData(rustCoreJwkData)
+            } catch (e: RustCoreException) {
+                throw Web5Exception.fromRustCore(e)
+            } catch (e: Exception) {
+                throw e
+            }
```
nitro-neal, line 26:
```suggestion
            try {
                val rustCoreJwkData = ed25519GeneratorGenerate()
                return Jwk.fromRustCoreJwkData(rustCoreJwkData)
            } catch (e: RustCoreException) {
                throw Web5Exception.fromRustCore(e)
            } 
```

bound/kt/src/main/kotlin/web5/sdk/crypto/Ed25519Generator.kt
nitro-neal: also since there are a lot of these it could be cleaner to do it this way with return try:
```
    return try {
        val rustCoreJwkData = ed25519GeneratorGenerate()
        Jwk.fromRustCoreJwkData(rustCoreJwkData)
    } catch (e: RustCoreException) {
        throw Web5Exception.fromRustCore(e)
    }
```

nitro-neal (approved):

nitro-neal:
added this issue to investigate further for less verbose exception handling  - https://github.com/TBD54566975/web5-rs/issues/335

PR #337, opened 2024-09-04
rust impl of vps

nitro-neal:


bindings/web5_uniffi_wrapper/src/errors.rs
```diff
@@ -2,7 +2,7 @@ use serde_json::Error as SerdeJsonError;
 use std::fmt::Debug;
 use thiserror::Error;
 use web5::credentials::presentation_definition::PexError;
-use web5::credentials::CredentialError;
+use web5::credentials::VerificationError;
```
nitro-neal, line 5: i changed the name so vp and vc verification can share the same errors

docs/API_DESIGN.md
```diff
@@ -88,9 +91,11 @@ CLASS VerifiableCredential
   PUBLIC DATA id: string
   PUBLIC DATA type: []string
   PUBLIC DATA issuer: Issuer
+  PUBLIC DATA credentialSubject: CredentialSubject
```
KendallWeihe, line 94: we don't really have a good solution for casing in our APID, but maybe this just to be consistent

```suggestion
  PUBLIC DATA credential_subject: CredentialSubject
```

KendallWeihe, line 94: we don't really have a good solution for casing in our APID, but maybe this just to be consistent

```suggestion
  PUBLIC DATA credential_subject: CredentialSubject
```

bindings/web5_uniffi_wrapper/src/errors.rs
KendallWeihe: initial reaction is skepticism, but after more consideration I agree

crates/web5/src/credentials/verifiable_presentation_1_1.rs
```diff
@@ -0,0 +1,498 @@
+use crate::credentials::josekit::{JoseSigner, JoseVerifier, JoseVerifierAlwaysTrue};
```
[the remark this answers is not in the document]
KendallWeihe: I'm kinda thinking we may want to do a little code reorganization to try and 1:1 match [the web5 spec](https://tbd54566975.github.io/web5-spec/), but not necessary at this moment, can consider later

crates/web5/src/credentials/mod.rs
```diff
@@ -14,13 +14,14 @@ pub use status_list_credential::{
     STATUS_LIST_CREDENTIAL_TYPE,
 };
 pub mod verifiable_credential_1_1;
+pub mod verifiable_presentation_1_1;
```
KendallWeihe, line 17: I think for both of these lines we should follow the same pattern of...

```suggestion
mod verifiable_credential_1_1;
mod verifiable_presentation_1_1;
```

and then subsequently follow-up with a `pub use` for the specific features which the APID makes available -- not strictly necessary at this time, but if not done here I'll probably follow-up with these proposed changes

bindings/web5_uniffi_wrapper/src/credentials/verifiable_presentation_1_1.rs
```diff
@@ -0,0 +1,76 @@
...
+use web5::credentials::verifiable_presentation_1_1::VerifiablePresentationCreateOptions as InnerVerifiablePresentationCreateOptions;
+
+#[derive(Default)]
+pub struct VerifiablePresentationCreateOptions {
+    pub id: Option<String>,
+    pub context: Option<Vec<String>>,
+    pub r#type: Option<Vec<String>>,
+    pub issuance_date: Option<SystemTime>,
+    pub expiration_date: Option<SystemTime>,
+}
```
KendallWeihe, line 13: you can actually just re-use the `web5::credentials::verifiable_presentation_1_1::VerifiablePresentationCreateOptions` here rather than re-defining the struct and mapping to/from. it's not the case with `VerifiableCredentialCreateOptions` because that struct has a property within it which has to be json-serialized over the FFI (evidence) whereas here you just have primitive types.

bindings/web5_uniffi_wrapper/src/credentials/verifiable_presentation_1_1.rs
```diff
@@ -0,0 +1,76 @@
...
+        bearer_did: Arc<BearerDid>,
+        verification_method_id: Option<String>,
+    ) -> Result<String> {
+        let vp_jwt = self.inner_vp.sign(&bearer_did.0, verification_method_id)?;
+        Ok(vp_jwt)
+    }
+}
+
+#[derive(Clone)]
+pub struct VerifiablePresentationData {
```
KendallWeihe, line 68: same thing here actually, these are all primitive types so you can re-use `InnerVerifiablePresentation` instead of re-defining the struct here

bound/kt/src/main/kotlin/web5/sdk/vc/VerifiablePresentation.kt
```diff
@@ -0,0 +1,107 @@
...
+    val type: List<String>,
+    val id: String,
+    val holder: String,
+    val issuanceDate: Date,
+    val expirationDate: Date? = null,
+    val verifiableCredential: List<String>,
+    internal val rustCoreVerifiablePresentation: RustCoreVerifiablePresentation,
+) {
+    companion object {
+        internal fun fromRustCore(rustCoreVerifiablePresentation: RustCoreVerifiablePresentation): VerifiablePresentation {
```
KendallWeihe, line 29: looks like this isn't used anywhere so it's not needed, right?

KendallWeihe (approved):
🏆 🥇 🙌

crates/web5/src/credentials/verifiable_presentation_1_1.rs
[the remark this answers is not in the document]
nitro-neal: image
maybe this is in error?

crates/web5/src/credentials/mod.rs
nitro-neal: yup added

bindings/web5_uniffi_wrapper/src/credentials/verifiable_presentation_1_1.rs
nitro-neal: I see.. yup makes sense

crates/web5/src/credentials/verifiable_presentation_1_1.rs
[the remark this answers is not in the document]
KendallWeihe: Screenshot 2024-09-05 at 12 24 42 PM

I believe it is (from [here](https://tbd54566975.github.io/web5-spec/#presentation-exchange))

cc @decentralgabe

bindings/web5_uniffi_wrapper/src/credentials/verifiable_presentation_1_1.rs
nitro-neal: I tried to get this to work but i think we need it?, will update if needed in future pr if its possible

PR #338, opened 2024-09-04
Update to latest web5-spec

KendallWeihe:
Discovered from [here](https://github.com/TBD54566975/tbdex-rs/actions/runs/10709579181/job/29694522139). We had this pointing to a feature branch which no longer exists. This is latest on main.


nitro-neal (approved):
nice

PR #339, opened 2024-09-04
Replace reqwest with hyper

diehuxx:
`reqwest` is causing some problems in https://github.com/TBD54566975/web5-rs/pull/319, so we're going to use a lower level HTTP library which won't have those same issues. Also, this is a good opportunity to refactor some functions as `async`, e.g. DID web and dht resolution. Unfortunately, testing async functions breaks the `UnitTestSuite` thing we have set up. We're due to replace that with something more robust anyway, so we can replace that in a follow up PR.


diehuxx:
Closing, because @KendallWeihe mentioned a desire to move away from any HTTP dependency and just use std lib

PR #340, opened 2024-09-05
Propose using web5 Result and .into()

KendallWeihe:


KendallWeihe:
there's no reason this is strictly necessary or better then the existing, so I'm closing this

PR #344, opened 2024-09-05
Security + Codecov CI updates

finn-tbd:
This brings the web5-rs security scanning in line with what most other repos use, and ensures that most updates can be applied automatically when the central repo is updated.


KendallWeihe (approved):
nice!

PR #346, opened 2024-09-09
Remove josekit

KendallWeihe:
- We're only using josekit for it's data structures
- We roll our own cryptography
- Negative externalities outweigh benefits
  - Esoteric code
  - No less overhead
  - OpenSSL!

The goal here is not to build generalized JOSE solutions, it's to build JOSE solutions relevant to web5


KendallWeihe:
Once we remove reqwest then both openssl and ring will be entirely removed from the dep chart, and thereafter we can remove the openssl build code from the two musl dockerfile's

nitro-neal (approved):
Amazing work, lots of good stuff here

crates/web5/src/lib.rs
```diff
@@ -2,9 +2,10 @@ pub mod credentials;
 pub mod crypto;
 pub mod dids;
 
+mod datetime;
 pub mod errors;
+pub mod jose;
```
diehuxx, line 7: i understand that we're trying to remove josekit as a dependency, but we didn't previous re-export josekit's API. why are we exposing a jose API now? i thought this PR is meant to change the web5 crate's implementation without changing the API

crates/web5/src/credentials/verifiable_credential_1_1.rs
```diff
@@ -262,13 +264,13 @@ mod tests {
...
...sSVRUbHRkSGx5Y0dsWWRGRlVNR3B4Wms1MmFXTm5RVGxCVkRnME1IWTFZMDh5YjFSckluMCMwIn0.eyJ2YyI6eyJAY29udGV4dCI6WyJodHRwczovL3d3dy53My5vcmcvMjAxOC9jcmVkZW50aWFscy92MSJdLCJpZCI6InVybjp1dWlkOjZmYTQ2MDVjLWFlZGItNGQ2NC05NzdiLTFmY2NmYTU1ZTM1ZiIsInR5cGUiOlsiVmVyaWZpYWJsZUNyZWRlbnRpYWwiXSwiaXNzdWVyIjoiZGlkOmp3azpleUpoYkdjaU9pSkZaREkxTlRFNUlpd2lhM1I1SWpvaVQwdFFJaXdpWTNKMklqb2lSV1F5TlRVeE9TSXNJbmdpT2lKSFZ6WkZURGxJVFRsdGRIbHljR2xZZEZGVU1HcHhaazUyYVdOblFUbEJWRGcwTUhZMVkwOHliMVJySW4wIzAiLCJpc3N1YW5jZURhdGUiOiIyMDI0LTA4LTI4VDEyOjQyOjI3Ljc3Mjg4OSswMDowMCIsImV4cGlyYXRpb25EYXRlIjpudWxsLCJjcmVkZW50aWFsU3ViamVjdCI6eyJpZCI6ImRpZDpkaHQ6cWdtbXB5anc1aHducWZnem43d21ybTMzYWR5OGdiOHo5aWRlaWI2bTlnajR5czZ3bnk4eSJ9fSwiaXNzIjoiZGlkOmp3azpleUpoYkdjaU9pSkZaREkxTlRFNUlpd2lhM1I1SWpvaVQwdFFJaXdpWTNKMklqb2lSV1F5TlRVeE9TSXNJbmdpT2lKSFZ6WkZURGxJVFRsdGRIbHljR2xZZEZGVU1HcHhaazUyYVdOblFUbEJWRGcwTUhZMVkwOHliMVJySW4wIzAiLCJqdGkiOiJ1cm46dXVpZDo2ZmE0NjA1Yy1hZWRiLTRkNjQtOTc3Yi0xZmNjZmE1NWUzNWYiLCJzdWIiOiJkaWQ6ZGh0OnFnbW1weWp3NWh3bnFmZ3puN3dtcm0zM2FkeThnYjh6OWlkZWliNm05Z2o0eXM2d255OHkiLCJuYmYiOjE3MjQ4NDg5NDcsImlhdCI6MTcyNDg0ODk0N30.-JwIGYZ9HlJASYxdRBWY5KlwP0iJUxWUOU6BsOR74VeC-zKgZb9WWZR08OVD-wv0X8KD5--0K5Dr9r5fL3B0Aw-invalid-signature"#;
```
diehuxx, line 265: why are we changing this test value? was it incorrect?

bound/kt/src/test/kotlin/web5/sdk/vc/VerifiableCredentialTest.kt
```diff
@@ -736,13 +736,13 @@ class VerifiableCredentialTest {
...
...ERsSVRUbHRkSGx5Y0dsWWRGRlVNR3B4Wms1MmFXTm5RVGxCVkRnME1IWTFZMDh5YjFSckluMCMwIn0.eyJ2YyI6eyJAY29udGV4dCI6WyJodHRwczovL3d3dy53My5vcmcvMjAxOC9jcmVkZW50aWFscy92MSJdLCJpZCI6InVybjp1dWlkOjZmYTQ2MDVjLWFlZGItNGQ2NC05NzdiLTFmY2NmYTU1ZTM1ZiIsInR5cGUiOlsiVmVyaWZpYWJsZUNyZWRlbnRpYWwiXSwiaXNzdWVyIjoiZGlkOmp3azpleUpoYkdjaU9pSkZaREkxTlRFNUlpd2lhM1I1SWpvaVQwdFFJaXdpWTNKMklqb2lSV1F5TlRVeE9TSXNJbmdpT2lKSFZ6WkZURGxJVFRsdGRIbHljR2xZZEZGVU1HcHhaazUyYVdOblFUbEJWRGcwTUhZMVkwOHliMVJySW4wIzAiLCJpc3N1YW5jZURhdGUiOiIyMDI0LTA4LTI4VDEyOjQyOjI3Ljc3Mjg4OSswMDowMCIsImV4cGlyYXRpb25EYXRlIjpudWxsLCJjcmVkZW50aWFsU3ViamVjdCI6eyJpZCI6ImRpZDpkaHQ6cWdtbXB5anc1aHducWZnem43d21ybTMzYWR5OGdiOHo5aWRlaWI2bTlnajR5czZ3bnk4eSJ9fSwiaXNzIjoiZGlkOmp3azpleUpoYkdjaU9pSkZaREkxTlRFNUlpd2lhM1I1SWpvaVQwdFFJaXdpWTNKMklqb2lSV1F5TlRVeE9TSXNJbmdpT2lKSFZ6WkZURGxJVFRsdGRIbHljR2xZZEZGVU1HcHhaazUyYVdOblFUbEJWRGcwTUhZMVkwOHliMVJySW4wIzAiLCJqdGkiOiJ1cm46dXVpZDo2ZmE0NjA1Yy1hZWRiLTRkNjQtOTc3Yi0xZmNjZmE1NWUzNWYiLCJzdWIiOiJkaWQ6ZGh0OnFnbW1weWp3NWh3bnFmZ3puN3dtcm0zM2FkeThnYjh6OWlkZWliNm05Z2o0eXM2d255OHkiLCJuYmYiOjE3MjQ4NDg5NDcsImlhdCI6MTcyNDg0ODk0N30.-JwIGYZ9HlJASYxdRBWY5KlwP0iJUxWUOU6BsOR74VeC-zKgZb9WWZR08OVD-wv0X8KD5--0K5Dr9r5fL3B0Aw-invalid-signature"
```
diehuxx, line 739: ditto https://github.com/TBD54566975/web5-rs/pull/346/files#r1752714076

crates/web5/src/credentials/sign.rs
```diff
@@ -1,23 +1,26 @@
...
 };
-use josekit::{jws::JwsHeader, jwt::JwtPayload};
-
-use super::{
-    josekit::JoseSigner, jwt_payload_vc::JwtPayloadVerifiableCredential,
-    verifiable_credential_1_1::VerifiableCredential,
-};
+use std::{collections::HashMap, time::SystemTime};
 
-pub fn sign_with_signer(
```
diehuxx, line 15: did we remove the functionality to sign a credential with an arbitrary Signer?

crates/web5/src/lib.rs
KendallWeihe: yeah, let's not quite make this public... yet. let's consider later after further deliberation. I think we probably will want to make it public, though that doesn't mean include it in the APID (though it may), primarily on two use cases: tbdex (which has a detached compact jws as it's message signatures) and the web5 cli (which, it's common place among devrel to want to inspect JWTs, especially when working with KCC matters). if we were to make this public, the messaging would be clear, that this is not intended to be a JOSE library we're building, but it _does_ offer JOSE features unique to web5 use cases (for example, web5 makes a requirement of the JOSE header kid being a DID VM id).

nevertheless, for now, I will not make it public

crates/web5/src/credentials/sign.rs
KendallWeihe: yeah, we weren't exposing it, and we are starting to bump up against poor design, because really we should probably add a `JoseSigner` which includes JOSE header values, otherwise we'd have to add an additional param here for the `Dsa` to be set as the `alg`

crates/web5/src/credentials/verifiable_credential_1_1.rs
KendallWeihe: this jwt was actually a secp256k1 which wasn't previously supported (until I just merged latest from main), but we didn't throw an error we just always assumed ed25519, so it was failing ed25519 because it was secp256k1

so I just changed it to ed25519 and actually created a bad signature as opposed to taking a good signature and appending on `-invalid-signature` (which is invalid for different reasons, which would be it's the wrong size as opposed to the right size but just an invalid value)

bound/kt/src/test/kotlin/web5/sdk/vc/VerifiableCredentialTest.kt
KendallWeihe: same reason, same test

KendallWeihe:
interestingly enough, even though we are hand-rolling our own jose here, just glossing over the code diff it seems to me like there's far more code removal in this PR than code additions... speaks loudly IMO

PR #350, opened 2024-09-10
Add secp256k1 signer & verifier, fix serde deserialize optional timestamp bug, add vc verify test vector

KendallWeihe:
~~We need to first merge this PR https://github.com/TBD54566975/web5-spec/pull/173 and then update our web5-spec commit~~

Edit: done

### DIFF
--- a/crates/web5/src/credentials/jwt_payload_vc.rs
+++ b/crates/web5/src/credentials/jwt_payload_vc.rs
@@ -22,13 +22,15 @@ pub struct JwtPayloadVerifiableCredential {
     #[serde(
         rename = "issuanceDate",
         serialize_with = "serialize_optional_system_time",
-        deserialize_with = "deserialize_optional_system_time"
+        deserialize_with = "deserialize_optional_system_time",
+        default
     )]
     pub issuance_date: Option<SystemTime>,
     #[serde(
         rename = "expirationDate",
         serialize_with = "serialize_optional_system_time",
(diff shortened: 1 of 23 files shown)


crates/web5/src/credentials/jwt_payload_vc.rs
```diff
@@ -22,13 +22,15 @@ pub struct JwtPayloadVerifiableCredential {
     #[serde(
         rename = "issuanceDate",
         serialize_with = "serialize_optional_system_time",
-        deserialize_with = "deserialize_optional_system_time"
+        deserialize_with = "deserialize_optional_system_time",
+        default
```
KendallWeihe, line 26: this was that serde deserialize bug -- turns out if you bring a custom deserializer then the `Option<T>` doesn't pick it up so you have to make it explicit

crates/web5/src/test_vectors.rs
```diff
@@ -324,4 +324,46 @@ mod test_vectors {
...
+        use crate::credentials::VerifiableCredential;
+
+        #[derive(Debug, serde::Deserialize)]
+        struct VerifyVectorInput {
+            #[serde(rename = "vcJwt")]
+            pub vc_jwt: String,
+        }
+
+        #[test]
+        fn verify() {
```
nitro-neal, line 339: aweomse!

nitro-neal (approved):
🥇 good stuff, looks good!

PR #352, opened 2024-09-11
Replace reqwest

KendallWeihe:


KendallWeihe:
Probably won't move forward with this approach, closing in favor of https://github.com/TBD54566975/web5-rs/tree/kendall/rm-reqwest

PR #354, opened 2024-09-12
Remove reqwest dependency

KendallWeihe:


KendallWeihe:
Converting to a draft for the time being for further consideration

diehuxx:
@KendallWeihe The failure is that we expect `Web5Error::Network("failed to lookup address information")` but instead we're getting `Web5Error::Network("No such host is known")`. All that differs is the text of the error message returned to us by `TcpStream::Connect` in `http.rs`.

I suspect that Windows rust implementation of `TcpStream::connect` returns one error message and Mac and Linux return another. In both cases, the error message conveys the same thing, i.e. This host isn't valid.

I'd be comfortable just updating the test to accept either error message. I added a commit doing this.

crates/web5/src/http.rs
```diff
@@ -0,0 +1,220 @@
...
+        )));
+    }
+
+    let json_value = serde_json::from_slice::<T>(&response.body)
+        .map_err(|err| Web5Error::Http(format!("unable to parse json response body {}", err)))?;
+
+    Ok(json_value)
+}
+
+pub fn get_bytes_as_http_response(url: &str) -> Result<HttpResponse> {
```
diehuxx, line 183: feel like we could just call this `get` and it would be less confusing. `get_bytes_as_http_response` sounds like i'm parsing the `url` as an http response

KendallWeihe: I agree! and then do the same for the `put_` function below too, yeah?

KendallWeihe:
On second thought, I want to spend more time further contemplating the implications here. Moving back to a draft for the time being.

KendallWeihe:
A leading alternative idea I have here: use `reqwest` inside the `web5` crate, but compile it out during binding, and over the FFI expose the ability to inject an http client lib into `web5` from the target languages native runtime. So in other words, for cross language bindings, we enable bring-your-own http client lib. This may be the best of both worlds; more overhead, but we compile away openssl/ring/etc so we reduce the byte code file size in the same way, and we're less likely to run into openssl bugs.

andresuribe87:
> A leading alternative idea I have here: use `reqwest` inside the `web5` crate, but compile it out during binding, and over the FFI expose the ability to inject an http client lib into `web5` from the target languages native runtime. So in other words, for cross language bindings, we enable bring-your-own http client lib. This may be the best of both worlds; more overhead, but we compile away openssl/ring/etc so we reduce the byte code file size in the same way, and we're less likely to run into openssl bugs.

FWIW, that's similar to the approach that ktor uses to support different platforms. See [here](https://ktor.io/docs/client-engines.html), for more details.

PR #358, opened 2024-09-17
Artifact Publishing

nitro-neal:
This change refactors the ci and release github actions to closely match tbdex-rs.


.github/workflows/ci.yml
```diff
@@ -100,10 +67,8 @@ jobs:
     name: Build aarch64-apple-darwin target
     steps:
       - uses: actions/checkout@v2
-      - name: Install Rust 1.74.0
-        run: rustup toolchain install 1.74.0
-      - name: Set Rust 1.74.0 as default
-        run: rustup default 1.74.0
```
KendallWeihe, line 106: let's keep the version hard set to 1.74.0 to all of these lib builds

.github/workflows/ci.yml
```diff
@@ -295,11 +219,7 @@ jobs:
         run: |
           # Version resolution: use provided
           if [ -n "${{ github.event.inputs.version }}" ]; then
-            if [[ "${{ github.event.inputs.version }}" != *-SNAPSHOT ]]; then
-              resolvedVersion="${{ github.event.inputs.version }}-SNAPSHOT"
-            else
-              resolvedVersion="${{ github.event.inputs.version }}"
-            fi
```
KendallWeihe, line 302: not sure why we ever changed this in tbdex-rs, but I think we should keep this code as it was, the ensure we never accidentally publish a version which doesn't have `-SNAPSHOT` suffix. thoughts?

crates/web5/src/jose.rs
```diff
@@ -62,6 +62,7 @@ pub struct JwtClaims {
 impl FromJson for JwtClaims {}
 impl ToJson for JwtClaims {}
 
+#[allow(dead_code)]
```
KendallWeihe, line 65: was linting or something throwing an error here?

nitro-neal: yup, should we change the lint config to not throw on this stuff?

.github/workflows/ci.yml
nitro-neal: added back in

crates/web5/src/jose.rs
KendallWeihe: I'm confused why it would be throwing this tho, because we use this in VC JWTs and VP JWTs

.github/workflows/release.yaml
```diff
@@ -71,6 +90,166 @@ jobs:
...
+      RELEASE_TAG: ${{ steps.set-version-and-tag.outputs.RELEASE_TAG }}
+      RELEASE_VERSION: ${{ steps.set-version-and-tag.outputs.RELEASE_VERSION }}
+
+    name: Create Git Tag
+    steps:
+      - id: checkout
+        uses: actions/checkout@v4
+        with:
+          submodules: true
+          token: ${{ secrets.GITHUB_TOKEN }}
```
leordev, line 106: it needs to be `TBD_RELEASE_GITHUB_PERSONAL_ACCESS_TOKEN`

because we are actually using our tbd release bot to commit with mvn deploy

leordev: @ALRubinger I just noticed that we need to add this repo to have permission to this org secret, can you please provide it for us?

.github/workflows/release.yaml
```diff
@@ -71,6 +90,166 @@ jobs:
           files: web5_cli-x86_64-unknown-linux-musl
           repo-token: ${{ secrets.GITHUB_TOKEN }}
```
leordev, line 92: All the above `build_cli*` jobs are not present in [tbdex-rs release pipeline](https://github.com/TBD54566975/tbdex-rs/blob/main/.github/workflows/release.yml#L15-L23). Is there a reason we need it for web5?

.github/workflows/release.yaml
```diff
@@ -71,6 +90,166 @@ jobs:
...
+          else
+              echo "Error: No credentials"
+              exit 1
+          fi
+        env:
+          SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
+          SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
+          SIGN_KEY_PASS: ${{ secrets.GPG_SECRET_PASSPHRASE }}
+          SIGN_KEY: ${{ secrets.GPG_SECRET_KEY }}
+
```
leordev, line 252: I think you also need to update this release to include similar versioning strategy that you added to dokka in tbdex-rs below: https://github.com/TBD54566975/tbdex-rs/pull/120/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R204-R206

leordev (approved):
Release pipeline lgtm, just a couple of notes that should be addressed before merging.

.github/workflows/ci.yml
```diff
@@ -339,7 +307,7 @@ jobs:
           name: x86_64-pc-windows-msvc-dll
           path: bound/kt/src/main/resources/
 
-      - name: Deploy Snapshot / Verify
+      - name: Build and Test Kotlin Project
```
leordev, line 310: I'd probably leave something related to the Snapshot in this title, because we are creating a snapshot in our artifactory

.github/workflows/release.yaml
nitro-neal: The cli only lives in web5-rs (and not tbdex) so we want to release a cli with every version

.github/workflows/release.yaml
nitro-neal: yup good call

crates/web5/src/jose.rs
nitro-neal: image

PR #360, opened 2024-09-19
remove circular dependency

nitro-neal:
image

I got these out of order, fixing


nitro-neal:
so test vectors will work after next pull - 
image

leordev (approved):
lgtm

PR #362, opened 2024-09-20
Update release workflow

finn-tbd:
This fixes a number of issues with the previous release workflow.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -20,62 +20,58 @@ jobs:
     with:
       version: ${{ github.event.inputs.version }}
 
-  build_cli_aarch64-apple-darwin:
+  build-cli-aarch64-apple-darwin:
     runs-on: macos-latest
-    name: aarch64-apple-darwin
+    name: Build CLI (aarch64-apple-darwin)
     steps:
       - uses: actions/checkout@v2
-      - name: Install Rust 1.74.0
-        run: rustup toolchain install 1.74.0
-      - name: Set Rust 1.74.0 as default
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -229,6 +231,7 @@ jobs:
           path: bound/kt/src/main/resources/
 
       - name: Deploy Release Version of Kotlin Project to Maven Central
+        if: ${{ env.SONATYPE_PASSWORD != '' }}
```
nitro-neal, line 234: we still need this?

.github/workflows/release.yaml
```diff
@@ -250,6 +253,69 @@ jobs:
           SIGN_KEY_PASS: ${{ secrets.GPG_SECRET_PASSPHRASE }}
           SIGN_KEY: ${{ secrets.GPG_SECRET_KEY }}
 
+      - name: Create GitHub Release
+        uses: actions/create-release@0cb9c9b65d5d1901c1f53e5e66eaf4afd303e70e # v1.1.4 final release
```
nitro-neal, line 257:
```suggestion
        uses: actions/create-release@0cb9c9b65d5d1901c1f53e5e66eaf4afd303e70e
```

nitro-neal (approved):

.github/workflows/release.yaml
finn-tbd: not really but it will be useful if we want to debug this in the future.

finn-tbd: Had to update some other stuff so I took this out too.

PR #363, opened 2024-09-20
Add wasm foundation

KendallWeihe:


bound/kt/pom.xml
```diff
@@ -8,7 +8,7 @@
   <!-- Artifact Information  -->
   <groupId>xyz.block</groupId>
   <artifactId>web5</artifactId>
-  <version>0.0.0-main-SNAPSHOT</version>
+  <version>0.0.0-main-kweihe-local-SNAPSHOT</version>
```
KendallWeihe, line 11: todo: undo this prior to merge

LICENSE
```diff
@@ -1,201 +0,0 @@
-                                 Apache License
```
KendallWeihe, line 1: todo: fix this before merging (accidental)

Justfile
```diff
@@ -8,9 +8,12 @@ setup:
     git submodule update --init --recursive
   fi
   if [[ "$(cargo 2>&1)" == *"rustup could not choose a version of cargo to run"* ]]; then
-    rustup default 1.74.0
+    rustup default 1.76.0 # TODO undo this
```
KendallWeihe, line 11: @diehuxx I know this is no bueno but thinking of leaving it for now. I had to do it because of the `wasm-pack` version we're using... though we could look into moving to a lower `wasm-pack` I'm not sure what the consequences would be. Lmk your thoughts.

bindings/web5_wasm/src/crypto/key_managers.rs
```diff
@@ -0,0 +1,75 @@
...
+    }
+}
+
+#[wasm_bindgen]
+pub fn new_in_memory_key_manager() -> Result<WasmKeyManager> {
+    Ok(InMemoryKeyManager::new().into())
+}
+
+#[wasm_bindgen]
+extern "C" {
```
KendallWeihe, line 48: here's an example of a _foreign_ implementation, for which we'll use this same pattern for a foreign `fetch` implementation

crates/web5/src/http.rs
```diff
@@ -0,0 +1,138 @@
+use crate::errors::{Result, Web5Error};
```
KendallWeihe, line 1: @diehuxx hold off on reviewing anything http-related until we have #354 merged, and I'll wrangle this around to use `http-std` as the default and enable developer-override

crates/web5/src/jose.rs
```diff
@@ -26,54 +26,19 @@ pub struct JoseHeader {
...
-    pub exp: Option<SystemTime>,
-    #[serde(flatten, skip_serializing_if = "Option::is_none")]
-    pub additional_properties: Option<HashMap<String, JsonValue>>,
-}
-
-impl FromJson for JwtClaims {}
-impl ToJson for JwtClaims {}
-
-pub struct Jwt {
+pub struct Jws {
```
KendallWeihe, line 29: oh btw @diehuxx @nitro-neal forgot to mention earlier, but I've also added `Jws` here b/c tbdex uses a detatched JWS for message signatures

Justfile
diehuxx: That's fine. We CI job that checks if the web5 crate has MSRV 1.74. The CI job checks `crates/web5/Cargo.toml` to get `rust-version = 1.74`, so as long as we don't change that, we're fine.

The only drawback of using a higher rust version during development is that we may accidentally use features in `crates/web5` that aren't supported in version 1.74. If that happens, we'll just notice them in the CI job failure and fix afterward.

bound/typescript/src/crypto/key_managers/index.ts
```diff
@@ -0,0 +1,35 @@
...
+  constructor(wasmKeyManager: wasm.WasmKeyManager) {
+    this.wasmKeyManager = wasmKeyManager
+  }
+
+  importPrivateJwk(privateJwk: Jwk): Jwk {
+    try {
+      const wasmJwk = this.wasmKeyManager.import_private_jwk(privateJwk.toWasmJwk());
+      return Jwk.fromWasmJwk(wasmJwk);
+    } catch (error) {
+      throw catchWeb5Error(error);
```
nitro-neal, line 23: its strange throwing a function, 

Chad says we can do this:
```
// Define the decorator
function CatchWeb5Error(
  target: Object,
  propertyKey: string,
  descriptor: PropertyDescriptor
) {
  const originalMethod = descriptor.value;

  descriptor.value = function (...args: any[]) {
    try {
      return originalMethod.apply(this, args);
    } catch (error) {
      throw catchWeb5Error(error);
    }
  };

  return descriptor;
}

// Apply the decorator to your methods
class YourClass {
  @CatchWeb5Error
  computeThumbprint(): string {
    const wasmJwk = this.toWasmJwk();
    return wasmJwk.compute_thumbprint();
  }

  @CatchWeb5Error
  anotherMethod(): void {
    // Your method implementation
  }
}
```

nitro-neal: We can think about it and if we want to go this way or another can make an issue to update error handling across the board

KendallWeihe: But yeah, in a follow up PR

crates/web5/src/credentials/sign.rs
```diff
@@ -38,6 +38,7 @@ pub fn sign_with_did(
     additional_properties.insert("vc".to_string(), vc_claim.to_json_value()?);
 
     let claims = JwtClaims {
+        aud: None,
```
nitro-neal, line 41: nice

nitro-neal (approved):

bound/typescript/src/crypto/key_managers/index.ts
KendallWeihe: looks like decorators are still considered an experimental feature so I may settle on a higher order function for now

crates/web5/src/lib.rs
```diff
@@ -4,8 +4,7 @@ pub mod dids;
 
 mod datetime;
 pub mod errors;
-mod http;
-mod jose;
+pub mod jose;
```
diehuxx, line 7: didn't catch this before merge, but why are we exposing `jose` now? IIRC we wanted to keep this as an implementation detail.

KendallWeihe: tbdex uses detached compact jws as it's signature field

PR #371, opened 2024-09-30
Bind Dids

nitro-neal:


bound/typescript/src/wasm/generated.d.ts
```diff
@@ -28,9 +28,136 @@ export function new_ed25519_signer(jwk: WasmJwk): WasmSigner;
 */
```
nitro-neal, line 28: Do I need to add this file to .gitignore?

nitro-neal:
I did not add bound/dids yet because I didn't know if we were thinking to go the generated-mappings approach or just do it manual

bound/typescript/src/wasm/generated.d.ts
KendallWeihe: I'm indifferent on this at this moment, I can't think of a reason one way or the other

KendallWeihe (approved):

PR #372, opened 2024-09-30
Add async for I/O calls

KendallWeihe:


bindings/web5_uniffi_wrapper/src/credentials/presentation_definition.rs
```diff
@@ -1,4 +1,5 @@
 use crate::errors::Result;
+use futures::executor::block_on;
```
diehuxx, line 2: why `futures::executor::block_on` vs alternatives like `tokio::task::spawn_blocking`?

nitro-neal:
image

diehuxx (approved):
broadly this looks good. any details we can clean up in subsequent PRs.

crates/http-std/src/client.rs
```diff
@@ -1,35 +1,54 @@
...
     Put,
 }
 
-impl ToString for Method {
-    fn to_string(&self) -> String {
-        match self {
-            Method::Get => "GET".to_string(),
-            Method::Post => "POST".to_string(),
-            Method::Put => "PUT".to_string(),
+impl fmt::Display for Method {
```
nitro-neal, line 32: did you mean to change the name?

nitro-neal (approved):

crates/http-std/src/client.rs
KendallWeihe: Yeah I learned that implementing `Display` is slightly preferable to implementing `ToString` because implementing `Display` also implements `ToString` plus you get `format!()` support

bindings/web5_uniffi_wrapper/src/credentials/presentation_definition.rs
KendallWeihe: `futures` has a slightly more streamlined error handling, whereas with tokio I would have to add a mapper into our error type, see the `unwrap` below

```rust
    pub fn resolve(uri: &str) -> Self {
        let inner_result = task::spawn_blocking(move || {
            tokio::runtime::Handle::current().block_on(InnerResolutionResult::resolve(uri))
        }).unwrap(); // Handle unwrap or any error handling as required.

        Self(inner_result)
    }
```

bindings/web5_uniffi_wrapper/Cargo.toml
```diff
@@ -7,6 +7,7 @@ repository.workspace = true
 license-file.workspace = true
 
 [dependencies]
+futures = "0.3.30"
```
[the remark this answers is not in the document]
KendallWeihe: indeed! good catch, we can improve this later

bindings/web5_uniffi_wrapper/src/credentials/presentation_definition.rs
KendallWeihe: Weird, I guess uniffi requires tokio? I could've sworn I tested with futures, but oh well, guess I have to go the tokio route

PR #377, opened 2024-10-01
Allow BYO http impl

diehuxx:
If you turn on the feature flag "http_reqwest", web5 crate will supply its own http implementation that uses `reqwest`.


crates/web5/src/http.rs
```diff
@@ -0,0 +1,172 @@
+
+use std::{collections::HashMap, sync::Arc};
+use async_trait::async_trait;
+use once_cell::sync::OnceCell;
+
+#[async_trait]
+pub trait HttpClient: Send + Sync {
```
KendallWeihe, line 7: I changed the `HttpClient` trait a week or so ago to this https://github.com/TBD54566975/web5-rs/blob/30d475edfd56685e45c2f17840d911c731341c97/crates/http-std/src/client.rs#L7 & this is how `tbdex-ts` currently integrates the fetch interface -- I propose we stick with the API design I linked to

I think my proposal for this PR is: 
- create a new `http` module here inside of `web5`
- copy/pasta all of the existing `http-std/src` code into that module
  - maybe rename a few things with the prefix `Http*`?
  - don't copy over the `default_client.rs` module (so basically two options: BYO or reqwest)
  - don't copy over the `error.rs` module use `TbdexError` instead of that existing `http-std::Error`
  - integrate the feature flags you've added in this PR into that code (in place of the existing wasm compiler conditionals)
- remove the `http-std` crate
- use `lazy_static` instead of `once_cell` since we already have `lazy_static` (or entirely replace `lazy_static` w/ `once_cell`? (the goal being, limit the number of dependencies)

diehuxx: Agreed with all except the last point.

> use lazy_static instead of once_cell since we already have lazy_static

`lazy_static` is good for cases where we can lazily evaluate a value, without needing to set it explicitly. For our purposes, the crate consumer wants to explicitly set the `HTTP_CLIENT`. `once_cell` is able to explicitly set it. See [this pattern](https://docs.rs/once_cell/1.20.1/once_cell/#safe-initialization-of-global-data) in the `once_cell` README.

> (the goal being, limit the number of dependencies)

`once_cell` is a small crate with no dependencies by default. `lazy_static` is the same way. Since they serve slightly different use cases in our codebase, and they introduce very little size to our , it's worth keeping both.

KendallWeihe:
Looks like you may have accidentally deleted `bound/typescript/package-lock.json`

crates/web5/src/http.rs
```diff
@@ -0,0 +1,222 @@
...
+pub enum Method {
+    Get,
+    Post,
+    Put,
+}
+
+pub struct HttpResponse {
+    pub status_code: u16,
+    pub headers: HashMap<String, String>,
+    pub body: Vec<u8>,
```
KendallWeihe, line 84: Can we change this to `pub body: Option<Vec<u8>>`?

crates/web5/src/dids/methods/did_web/resolver.rs
```diff
@@ -43,10 +48,20 @@ impl Resolver {
     }
 
     pub async fn resolve(&self) -> Result<ResolutionResult, ResolutionMetadataError> {
-        let response = http_std::fetch(&self.http_url, None)
+        let headers: HashMap<String, String> = HashMap::from([
+            ("Host".to_string(), "{}".to_string()),
+            ("Connection".to_string(), "close".to_string()),
+            ("Accept".to_string(), "application/json".to_string()),
+        ]);
```
KendallWeihe, line 55: do you think all of these headers are necessary? I've noticed you have added them throughout the various http calls. I'm guessing it's slightly more proper this way? just checking this isn'y copy/pasta from our roll-our-own http client code

crates/web5/src/lib.rs
```diff
@@ -4,8 +4,25 @@ pub mod dids;
 
 mod datetime;
 pub mod errors;
+mod http;
 pub mod jose;
 pub mod json;
 
+pub use http::set_http_client;
```
KendallWeihe, line 11: don't we also need to expose the `HttpClient` trait in order for developers to bring their own?

IMO we should simply `pub mod http;` everything, otherwise in `tbdex-rs` we're going to have to copy/pasta all the http code for it's own BYO functionality (which over there is where the hard requirement kicks in because we have to implement a foreign fetch over the wasm ffi)

PR #390, opened 2024-10-09
Support did_dht_resolution in C

andresuribe87:
## What type of PR is this? (check all applicable)

- [ ] ♻️ Refactor
- [x] ✨ New Feature
- [ ] 🐛 Bug Fix
- [x] 📝 Documentation Update
- [ ] 👷 Example Application
- [ ] 🧑‍💻 Code Snippet
- [ ] 🎨 Design
- [ ] 📖 Content
- [ ] 🧪 Tests
- [ ] 🔖 Release
- [ ] 🚩 Other

## Description

Adds support for `did_dht_resolution` in a C ABI.

Let me know if it's cool to include the tokio runtime `web5_c` library that's being built by the `bindc` recipe. 

## Related Tickets & Documents

N/A - couldn't find any. I wanted to add this because I'm aiming to extend curl so that it can handle `did` protocols. This is critical in order to promote interop. Since curl is written in C, this is a simple first step to use the rust-core to power the curl fork I'll be working on.

## Mobile & Desktop Screenshots/Recordings

N/A

## Added code snippets?
- [x] 👍 yes
- [ ] 🙅 no, because they aren't needed

## Added tests?

- [ ] 👍 yes
- [x] 🙅 no, because they aren't needed
- [ ] 🙋 no, because I need help

### No tests? Add a note

The provided example code is the test itself. There is a README with the explanation. 

## Added to documentation?

- [x] 📜 readme
- [ ] 📜 contributing.md
- [ ] 📓 general documentation
- [ ] 🙅 no documentation needed

### No docs? Add a note

Added a README that explains how to run the given code.


bindings/web5_c/src/c.rs
```diff
@@ -10,6 +10,52 @@ pub extern "C" fn free_string(s: *mut c_char) {
...
+        Ok(s) => s,
+        Err(_) => return ptr::null_mut(),
+    };
+
+    let gateway_url = if gateway_url.is_null() {
+        None
+    } else {
+        match unsafe { CStr::from_ptr(gateway_url).to_str() } {
+            Ok(s) => Some(s.to_string()),
+            Err(_) => return ptr::null_mut(),
```
KendallWeihe, line 39: at some point we should introduce better error handling over the C FFI, but it's not necessary at this moment

examples/CExample/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+3. Generate the C dynamic library by running:
+
+   ```
+   just bindc
+   ```
+
+   This command will build the necessary C bindings for the Web5 SDK.
+
+4. Navigate to the `examples/CExample` directory where the example-specific Justfile is located.
```
KendallWeihe, line 27: this step is a little non-sensical since we're already within the `examples/CExample` directory

examples/CExample/README.md
```diff
@@ -0,0 +1,51 @@
...
+2. Printing the resolved DID document or an error message
+
+This example demonstrates how to resolve a DHT-based DID using the Web5 SDK in C. It uses the `did_dht_resolve` function with a hardcoded DID URI and the default gateway. The program then prints either the resolved DID document or an error message to the console.
+
+
+## Running the Example
+
+To run this C example:
+
+1. Ensure you have a C compiler installed (e.g. GCC) and [Just](https://github.com/casey/just) command runner.
```
KendallWeihe, line 15: fwiw we already have `just` included [in the hermit dependencies](https://github.com/TBD54566975/web5-rs/blob/main/bin/.just-1.23.0.pkg), though I'd be curious if hermit has a `gcc` dependency available b/c if not then the developer will have to bring-their own which would be the first of it's kind in this repo

examples/CExample/README.md
KendallWeihe: oh I see, but `just bindc` is at the root level of the project, never mind this makes sense

KendallWeihe (approved):
well done 👏

examples/CExample/README.md
andresuribe87: > fwiw we already have just included in the hermit dependencies,

yay thanks didn't know that. I'll update accordingly. 

> though I'd be curious if hermit has a gcc dependency available b/c if not then the developer will have to bring-their own which would be the first of it's kind in this repo

when I checked, I looked only at https://cashapp.github.io/hermit/sdks/ and didn't see any C SDKs. Upon closer inspection, I see that there is `macos-cross-gcc-aarch64-unknown-linux-musl` is available. I'm concerned that it is macOS only and is meant for cross-compilation. What are your thoughts on this?

KendallWeihe: Yeah I think sticking with byo-gcc is fine, so long as it's called out

bindings/web5_c/src/c.rs
andresuribe87: Created https://github.com/TBD54566975/web5-rs/issues/392 to track.

PR #396, opened 2024-10-16
fix outside pr comment

nitro-neal:
Chad says:

Explanation:

github.event_name == 'pull_request': Checks if the event that triggered the workflow is a pull request.
github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name: Compares the repository names of the head (source) and base (target) of the pull request. If they are the same, it means the PR is from an inside account.
comment-on-pr: This input is now dynamically set to true only when the PR is from an inside account. Otherwise, it defaults to false.
This change ensures that the action will only attempt to comment on pull requests from inside accounts, avoiding permission issues or unwanted comments on PRs from outside contributors.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -309,7 +309,7 @@ jobs:
           vector-regex: (\w+)
           fail-on-missing-vectors: false
           fail-on-failed-test-cases: true
-          comment-on-pr: true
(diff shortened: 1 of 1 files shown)


leordev (approved):
🎉 great work Chad!!!

PR #405, opened 2024-10-31
Removing Hacktoberfest related actions

blackgirlbytes:
Since Hacktoberfest is ending, I'm removing/cleaning up actions and templates related to Hactkoberefst


KendallWeihe (approved):
The day after Halloween was always more sad to me than the day after Christmas 😢

PR #406, opened 2024-12-27
Implement portable DID conversion and key export functionality

botshelomokoka:


botshelomokoka (commented):
Adds WASM functions to convert between a Bearer DID and a portable DID and supplies a method to retrieve a signer. This improves WASM bindings by providing a consistent interface for portable DID handling and signing.

